PR #2, opened 2017-09-15
added support for handling string response body

amarzavery:


lib/util/utils.ts
```diff
@@ -313,10 +313,23 @@ export async function dispatchRequest(options: WebResource): Promise<HttpOperati
...
+      const textResponse = operationResponse.bodyAsText;
+      // As per, http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.25 and
+      // https://spacetelescope.github.io/understanding-json-schema/reference/type.html JSON has
+      //  ("null", "boolean", "object", "array", "number", or "string"), or "integer" as basic types.
+      // For JSON.parse() to a parse a string, the input should have double quotes in it like "\"Some Text\"".
+      // Since this is not the case, we want to make sure that the response body was actually a string
+      // and not a malformed JSON object or JSON array. If the response body was null like "null" or
+      // a number like "10" or a boolean like "true" or "false", then JSON.parse() handles those
+      // scenarios correctly. If the input does not start with "{" or "[" then it is most probably a string.
+      if (!(textResponse.startsWith("{") || textResponse.startsWith("["))) {
```
amarzavery: this condition will pass if someone sent "true1". So simply checking for a string starting with "{" or "[" will not help. We need to do better. May be checking for content-type. If content-type is text/json then we absorb the error and let the string pass as-is would be a better solution.

PR #6, opened 2017-10-17
replacing eval by traversing recursively in the object.

amarzavery:


lib/serializer.ts
```diff
@@ -435,20 +435,16 @@ export class Serializer {
...
           let propertyInstance;
-          try {
-            /*jslint evil: true */
-            propertyInstance = eval(jpath.join(""));
-          } catch (err) {
-            continue;
-          }
+          let res = responseBody;
+          // traversing the object step by step.
+          paths.forEach((item) => {
```
timotheeguerin: would maybe just `for(const item of paths)` instead of forEach as its recommended I think for es6 and more performant

PR #11, opened 2018-01-24
make isNode more organic rather than using some package

### LINKED ISSUE #10: The module dependency 'os' does not exist when used in React Native
I'm using AutoRest to generate a typescript rest client and converting that to ES6 to be used in a React Native app. The generated client relies on the 'ms-rest-js' runtime which relies on the nodejs module 'os' to hydrate the operating system info on the user agent string in the MsRestUserAgentFilter. Unfortunately the 'os' module is not implemented in React Native as far as I'm aware. 

I can pull in the nodejs 'os' module to get the bundler to succeed but unfortunately 'detect-node' returns true when running within React Native. This causes code to execute against the 'os' module but it's not implemented within React Native which yields a runtime error at line 23 of msRestUserAgentFilter.ts:
`const osInfo = '(${os.arch()}-${os.type()}-${os.release()})';`

If I use a debugger and set IsNode to false, everything works as expected.

When I manually replace the module dependency 'os' with '[react-native-os'](https://github.com/PeelTechnologies/react-native-os) in this scenario, everything behaves as expected. I spent a few hours attempting to alias require('os') to require('react-native-os') in my build but couldn't get it working. It's also hacky.

React Native may fall outside of intended use cases, but I'd be much obliged if 'ms-rest-js' (or an alternative package) could work in React Native without modification on my end, either by relying on the appropriate dependency ('react-native-os' vs 'os'), or by optionally bypassing/overriding the operating system info on the user agent string.

screen shot 2018-01-23 at 4 59 44 pm

amarzavery:
fixes #10


lib/util/utils.ts
```diff
@@ -28,6 +28,11 @@ export function getFetch(): Function {
  */
 export const myFetch = getFetch();
 
+/**
+ * A constant that indicates whether the environment is node.js or browser based.
+ */
+export const isNode = typeof navigator === 'undefined' && typeof process !== 'undefined'
```
amarzavery: @geocachingsyntaxerror - This is the logic that I found from a package named [detect-browser](https://github.com/DamonOehlman/detect-browser/blob/master/index.js#L69) which is used to detect whether the app is in a node.js environment or not. I did a basic test in the node REPL and loading one of the samples from this repo in the browser. It worked as expected in both the environments. 

Can you please pull the changes from detect branch of my fork of ms-rest-js and let me know if this works for react-native as well?

amarzavery: FYI, If this change does not work then we can try one more package [is-in-browser](https://www.npmjs.com/package/is-in-browser).

geocachingsyntaxerror: I pulled your detect branch and it worked as expected in react-native; isNode returns false as expected. Thank you.
detect

amarzavery:
@geocachingsyntaxerror - 0.2.4 version of ms-rest-js has been published with the fix. Thank you for using the autorest generated TS client. Feel free to open issues if you see any other problem.

Issue #15, opened 2018-02-01
Firefox 45 GET or HEAD has content body


thecodejedi:
Firefox 45 treats `body === null` differently from `body === undefined` 

ms-rest-js setting the body to null leads to ta TypeError for GET and HEAD requests as it has to be undefined for firefox.

```
TypeError: HEAD or GET Request cannot have a body.
Stack trace:
zoneify/<@webpack-internal:///../../../../zone.js/dist/zone.js:1022:33
dispatchRequest/</<@webpack-internal:///../../../../ms-rest-js/dist/lib/util/utils.js:335:42
step@webpack-internal:///../../../../ms-rest-js/dist/lib/util/utils.js:34:18
verb/<@webpack-internal:///../../../../ms-rest-js/dist/lib/util/utils.js:15:53
__awaiter</<@webpack-internal:///../../../../ms-rest-js/dist/lib/util/utils.js:9:15
ZoneAwarePromise@webpack-internal:///../../../../zone.js/dist/zone.js:875:29
__awaiter<@webpack-internal:///../../../../ms-rest-js/dist/lib/util/utils.js:5:1
dispatchRequest@webpack-internal:///../../../../ms-rest-js/dist/lib/util/utils.js:293:1
RequestPipeline.prototype.requestSink@webpack-internal:///../../../../ms-rest-js/dist/lib/requestPipeline.js:47:16
```

amarzavery:
I think fixing it in the runtime (ms-rest-js) is not the right way of doing it. We set the request body to null in the generator over [here](https://github.com/Azure/autorest.typescript/blob/master/src/vanilla/Templates/MethodTemplate.cshtml#L120). That would be a better place to change.
Request.body has [any](https://github.com/Azure/ms-rest-js/blob/master/lib/webResource.ts#L19) type. So not initializing requestContent to null and simply declaring it should be ok.

> However, the major issue that I would want to check is that, is it ok for request body to be set to undefined as the default instead of null?

@thecodejedi Can you make that change in the current PR in the autorest.typescript codegenerator?

PR #16, opened 2018-02-01
Supporting Firefox 45 GET/HEAD Requests

thecodejedi:
Fix for #15


thecodejedi:
Not required anymore.

Fixed in generator with Azure/autorest.typescript#20

PR #18, opened 2018-03-23
Duration fix

RikkiGibson:
This allows us to tolerate usage of multiple moment versions simultaneously. Dupes work already done in Azure/azure-sdk-for-node#2453. Suggest particularly looking at the first two commits--the third is more "noise" (sourcemaps, js files)


lib/serializer.ts
```diff
@@ -232,7 +232,7 @@ export class Serializer {
         }
         value = this.dateToUnixTime(value);
       } else if (typeName.match(/^TimeSpan$/ig) !== null) {
-        if (!isDuration(value)) {
+        if (!(isDuration(value) || (value.constructor && value.constructor.name === "Duration" && (value as any).isValid && (value as any).isValid()))) {
```
daschult: Do you need to check if isValid is a function? I'm sure you're fine not checking, but it isn't a function, will this throw an exception when you try to invoke it?

RikkiGibson: Can go with `typeof (value as any).isValid === 'function'`

RikkiGibson: I wish the moment type definitions simply included this member on Durations, maybe will just augment the type definition here locally

RikkiGibson: Done @daschult

daschult: Cool. I'm glad that worked out well.

Issue #9, opened 2017-12-18
Make sure to port the new changes done in ms-rest off lately to ms-rest-js as well.


amarzavery:
Make sure that the changes in ms-rest from [2.2.4](https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest/Changelog.md#224-10252017) onwards are reflected in the TS runtime.

PR #19, opened 2018-03-29
XML support

RikkiGibson:
- Adds xml2js
- Adds metadata to mappers for XML serialization (name, elementName, isWrapped, etc)
- Adds extra parsing behaviors for booleans and numbers. It's ambiguous to xml2js what kind of value conversion should take place, so we have to do it.
  - xml2js has "attribute value converters" for numbers and booleans that can be applied out of the box, but I would prefer to only do it if the Swagger says it should happen.

To do:
- [ ] ~~Handle model flattening~~ (seems unreasonably difficult because the AutoRest SerializedName doesn't take the XmlName into account)
  - PUT storage account with attributes, child elements, wrapped lists, embedded lists
- [x] Handle all review feedback asks
- [x] xml2js attrkey
- [x] port latest from azure sdk for node #9


lib/serializer.ts
```diff
@@ -310,38 +311,58 @@ export class Serializer {
         }
       }
 
-      for (const key in modelProps) {
-        if (modelProps.hasOwnProperty(key)) {
-          const paths = this.splitSerializeName(modelProps[key].serializedName);
-          const propName = paths.pop();
+      for (const key of Object.keys(modelProps)) {
```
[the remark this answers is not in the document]
RikkiGibson:
```js
for (const key of Object.keys(modelProps)) { ... }
```
is equivalent to
```js
for (const key in modelProps) {
  if (modelProps.hasOwnProperty(key) {
    ...
  }
}
```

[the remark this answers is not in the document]
daschult: Oh, weird. Okay.

lib/serializer.ts
```diff
@@ -310,38 +311,58 @@ export class Serializer {
...
 
-          let parentObject: any = payload;
           for (const pathName of paths) {
             const childObject = parentObject[pathName];
             if ((childObject === null || childObject === undefined) && (object[key] !== null && object[key] !== undefined)) {
               parentObject[pathName] = {};
             }
             parentObject = parentObject[pathName];
           }
+        }
```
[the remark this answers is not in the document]
RikkiGibson: Mainly I struggled to decide what parameter flattening means in XML. Maybe we can reason that out, or decide whether we're even supporting the feature or not.

lib/serializer.ts
```diff
@@ -512,13 +558,46 @@ export class Serializer {
...
   deserialize(mapper: Mapper, responseBody: any, objectName: string): any {
-    if (responseBody === null || responseBody === undefined) return responseBody;
+    if (responseBody == undefined) {
+      if (this.isXML && mapper.type.name === "Sequence" && !mapper.xmlIsWrapped) {
+        // Edge case for empty XML non-wrapped lists. xml2js can't distinguish
+        // between the list being empty versus being missing,
+        // so let's do the more user-friendly thing and return an empty list.
+        responseBody = [];
+      } else {
+        return responseBody;
```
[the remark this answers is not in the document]
RikkiGibson: So, this is a reasonable code structure question. This used to be a central "if the thing ain't there, don't try to map it" point. But there's now an edge case just above it that makes it a little less clear. I first tried to work on this by making every deserialization "helper" be resilient to undefined/null values (currently some are and others are not), but it was so much more copy-pasting of this kind of check that I eventually just went with this.

lib/serializer.ts
```diff
@@ -310,38 +311,58 @@ export class Serializer {
...
+        let propName: string | undefined;
+        let parentObject: any = payload;
+        if (this.isXML) {
+          if (propertyMapper.xmlIsWrapped) {
+            propName = propertyMapper.xmlName;
+          } else {
+            propName = propertyMapper.xmlElementName || propertyMapper.xmlName;
+          }
+        } else {
+          const paths = this.splitSerializeName(propertyMapper.serializedName);
```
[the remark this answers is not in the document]
RikkiGibson: Good question--looks like Dan wondered the same thing. I think I just glossed over it because the way the property name is obtained for XML is different than in JSON. We could maybe just take the propName extracted on line 323, and pass it to splitSerializeName, and carry on from there.

Note that the C# runtime does not handle parameter flattening for XML and we aren't aware of any services that would need it for XML.

lib/serializer.ts
```diff
@@ -512,13 +558,46 @@ export class Serializer {
    * @returns {object|string|Array|number|boolean|Date|stream} A valid deserialized Javascript object
    */
   deserialize(mapper: Mapper, responseBody: any, objectName: string): any {
-    if (responseBody === null || responseBody === undefined) return responseBody;
+    if (responseBody == undefined) {
```
amarzavery: === is different than ==

amarzavery: and why remove the responseBody === null check?

RikkiGibson: This is an intentional style change based on `foo === undefined || foo === null` being equivalent to `foo == undefined`.

Your input on which style to go with is welcomed--we can configure tslint to only allow `==` in the case of `expr == undefined` comparison.

lib/serializer.ts
```diff
@@ -512,13 +558,46 @@ export class Serializer {
...
+      if (this.isXML && mapper.type.name === "Sequence" && !mapper.xmlIsWrapped) {
+        // Edge case for empty XML non-wrapped lists. xml2js can't distinguish
+        // between the list being empty versus being missing,
+        // so let's do the more user-friendly thing and return an empty list.
+        responseBody = [];
+      } else {
+        return responseBody;
+      }
```
amarzavery: Isn't the else part redundant ? The outer return responseBody will eventually happen any ways.

lib/serializer.ts
amarzavery: By not having the null check this means that we want to process the responseBody when it is null. Is there are reason to do so?

lib/serializer.ts
RikkiGibson: I think Dan brought this up, see https://github.com/Azure/ms-rest-js/pull/19#discussion_r178192141

lib/serializer.ts
amarzavery: ok learnt something new today. Played in the node REPL.
```
> null == undefined
true
> null === undefined
false
>
```

lib/serializer.ts
```diff
@@ -435,8 +457,30 @@ export class Serializer {
         }
       }
 
-      for (const key in modelProps) {
-        if (modelProps.hasOwnProperty(key)) {
+      for (const key of Object.keys(modelProps)) {
+        const propertyMapper = modelProps[key];
+        let propertyObjectName = objectName;
+        if (propertyMapper.serializedName !== "") {
+          propertyObjectName = objectName + "." + modelProps[key].serializedName;
```
amarzavery: modelProps[key] -> propertyMapper (just for consistency)

lib/serializer.ts
```diff
@@ -5,11 +5,12 @@ import * as utils from "./util/utils";
 import { duration, isDuration } from "moment";
 const isBuffer = require("is-buffer");
 import * as isStream from "is-stream";
+import { isArray } from "util";
```
amarzavery: We don't need this. Be vary of importing the util library from node.js This has a direct impact on the bundle size in the browser. There is a function on the Array type `Array.isArray()`, use that instead.

lib/serializer.ts
```diff
@@ -310,38 +311,58 @@ export class Serializer {
...
-            if (object[key] === null || object[key] === undefined) {
-              throw new Error(`${key}" cannot be null or undefined in "${objectName}".`);
-            }
-          }
-          // make sure that readOnly properties are not sent on the wire
-          if (modelProps[key].readOnly) {
-            continue;
+        // make sure required properties of the CompositeType are present
+        if (modelProps[key].required && !modelProps[key].isConstant) {
+          if (object[key] === null || object[key] === undefined) {
```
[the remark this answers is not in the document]
RikkiGibson: Looking at it a bit more, I'm less enthused about the code churn, but could consider either changing all usages of `foo === null || foo === undefined` to `foo == undefined` in a different PR. Just don't want to create noise with that.

lib/serializer.ts
```diff
@@ -416,7 +437,8 @@ export class Serializer {
         name: "Composite"
       }
     };
-    if (responseBody !== null && responseBody !== undefined) {
+    responseBody = responseBody || {};
+    if (responseBody) {
```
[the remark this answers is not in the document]
RikkiGibson: Done

lib/serializer.ts
RikkiGibson: Done

lib/serializer.ts
RikkiGibson: Looking into removing all the "util" usages--would really like a promisify alternative though.

RikkiGibson: Looks like it's simply `new Promise(function(resolve, reject) { ...})` :)

amarzavery: We have never used node.js's util library at all. 
Note: We have a util folder in the runtime and importing stuff from there is fine. Just making sure there is no confusion.
And why would we ever need a Promisified version of `Array.isArray()` ?

amarzavery: `Array.isArray()` is a synchronous call.

RikkiGibson: I was making use of `promisify` from within the util module in another file :)

amarzavery: Ahh. I see. Yup that is super simple.
```
return new Promise((resolve, reject) => {
  if (err) reject(err);

  resolve(result);
});
```

That is what we are essentially doing in the generated code when we say we support callback and promises both [for example](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/storageManagement2/lib/operations/storageAccounts.js#L2994-L3000).

amarzavery: It is better to use arrow functions `() => {}` over `function () {}` since the `this` is preserved inside the arrow function

```javascript
function foo() {
  this.name = "foo";
   const bar = () => {
      console.log(this.name); // foo
   };
}
```

amarzavery: v/s 
```javascript
function foo() {
  this.name = "foo";
  let self = this;
   function bar() => {
      console.log(this.name); // undefined
      console.log(self.name); //"foo"
   };
}
```

lib/httpOperationResponse.ts
```diff
@@ -22,11 +22,12 @@ export class HttpOperationResponse {
...
    */
-  bodyAsText: string | null;
+  bodyAsText?: string | null;
+
   /**
-   * The response body as parsed JSON
+   * The response body as parsed JSON or XML
    */
-  bodyAsJson: { [key: string]: any } | Array<any> | string | number | boolean | null | void;
+  bodyAsJson?: { [key: string]: any } | Array<any> | string | number | boolean | null | void;
```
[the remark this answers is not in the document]
RikkiGibson: This also requires a change in ms-rest-azure-js, so I'll open a PR there pretty soon.

package.json
```diff
@@ -30,29 +30,31 @@
   "license": "MIT",
```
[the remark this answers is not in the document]
RikkiGibson: Bumped to 0.3.0

RikkiGibson:
I think I'm missing a bit of test coverage: an element with attributes and a simple text body like the following:

```xml
<pear color="green">Bartlett</pear>
```
I actually don't even know how to represent this data in a Swagger! An element with an attribute typically corresponds to an object type, where the property "color" would have "xmlIsAttribute = true", but there is no property name you can keep the text content under.

@lmazuel @olydis any idea how this would be represented?

olydis:
@RikkiGibson that's not representable for the very reason you state AFAIK - that was also one of the things we stumbled across for storage, I guess without an extension (I don't know, maybe to mark a certain property as the "body"? But that implies that there can be no other non-attribute properties... ughhhhh) that wouldn't work. But nothing in that direction happened from our side, do you see a need for that?

RikkiGibson:
Seems like this is a limitation of [OpenAPI 2.0 itself](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#xmlObject) hitting us and we should just let it lie until an important enough service starts asking for it.

RikkiGibson:
Addressed review feedback-- I'm shelving XML flattening for now because the serializedName doesn't give us the XmlNames that we need to traverse the parsed document.

Issue #8, opened 2017-12-18
Figureout how to serialize formdata with 'application/x-www-form-urlencoded' and 'multipart/form-data'


amarzavery:
- This is what the [request library](https://github.com/request/request#forms) does. They have separate fields `"form"` and `"formData"` that need to be used based on the `"Content-Type"` in the request header.
We have added a formDataFilter to the request pipeline in [ms-rest (node sdk)](https://github.com/Azure/azure-sdk-for-node/pull/2366/files) to solve this issue.
- We are depending on the `"fetch"` library which does not have good support for this.      
   - https://github.com/github/fetch/issues/263 and 
   - https://github.com/matthew-andrews/isomorphic-fetch/issues/30 
The above 2 issues should help us in finding the right solution.

RikkiGibson:
@amarzavery is this still an issue?

amarzavery:
In that case we will need to use some form-data package. Usually form-data is simple key/value pair kind of object. However the difficult part is calculating the boundary value and this value needs to be set in the header (I guess along with content-type value). Thus using a standard package in node.js that does that is beneficial for us.

angurusw:
dfsdfdsfsdaf

justinwilaby:
@amarzavery - 

Have you considered node-fetch with form-data? The isomorphic implementation can be either of 2 ways: 

## 1. Using Globals and the DefinePlugin for webpack
```js
/**
 * Node compatibility - This strategy hides
 * node-only dependencies from being bundled
 * in a browser based project even for dev builds.
 */
declare var __NODE__: boolean;
if (__NODE__) { // code is unreachable for __NODE__ == false and is stripped out
  (new Function(`
  if (!('FormData' in global)) {
    global.FormData = require('form-data');
  }
  if (!('fetch' in global)) {
    global.fetch = require('node-fetch');
  }`))();
}
```

## 2. Bring your own fetch and FormData
```js
HTTPClient.fetch = isNode() ? require('node-fetch') ? window.fetch
HTTPClient.FormData = isNode() ? require('form-data') ? window.FormData;
```

The first method does not require explicit configuration by the implementor but uses globals and could theoretically cause conflicts however unlikely they may be.

The second method requires that statics be configured with fetch and FormData by the implementor but allow more flexibility to provide their own fetch implementation on Node. Obviously, this is a no-op for most browser implementations.

Proxy support for Node users would then be provided *similar* to this regardless of the above implementation:

```js
[...]

PR #21, opened 2018-04-02
Updates from node SDK

RikkiGibson:
Applies updates based on the [Changelog](https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest/Changelog.md) in azure-sdk-for-node starting at version 2.2.4.

Some of the changes had already been ported over, such as 2.3.2 and 2.3.1.


lib/credentials/apiKeyCredentials.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.inQuery = options.inQuery;
+  }
+
+  /**
+   * Signs a request with the values provided in the inHeader and inQuery parameter.
+   *
+   * @param {WebResource} The WebResource to be signed.
+   * @returns {Promise<WebResource>} - The signed request object.
+   */
+  signRequest(webResource: WebResource) {
```
amarzavery:
```javascript
signRequest(webResource: WebResource): Promise<WebResource> {
}
```

lib/webResource.ts
```diff
@@ -271,10 +286,11 @@ export class WebResource {
 export interface RequestPrepareOptions {
   method: HttpMethods;
   url?: string;
-  queryParameters?: { [key: string]: any } | ParameterValue;
+  queryParameters?: { [key: string]: string | ParameterValue };
```
daschult: This is a pretty significant change (Changing from a type-less dictionary or a Parameter value to a dictionary with either string or ParameterValue values). Is it what you meant to do?

lib/webResource.ts
```diff
@@ -167,11 +168,22 @@ export class WebResource {
             }
           }
         }
-      }// end-of-for
+      } // end-of-for
       // append the queryString
       this.url += queryParams.join("&");
     }
 
+    // set formData parameters for 'application/x-www-form-urlencoded' or 'multipart/form-data'.
```
amarzavery: This is true only for the request.js library. https://github.com/request/request#requestoptions-callback. and search for `application/x-www-form-urlencoded`. You would see that the request library takes form and formData as separate options. Hence I made this change in the node sdk's ms-rest. Here we are using browser's fetch or node-fetch and we need to find out what the underlying library does when content-type is `"application/x-www-form-urlencoded"` v/s `"multipart/form-data"`.

lib/webResource.ts
amarzavery: How about this `string | number | boolean | object | array | ParameterValue`?

amarzavery: This would still make sure that null, undefined will not get in. However it will let the user know what he/she can provide.

lib/webResource.ts
RikkiGibson: So basically setting a `form` property on the request object was just because you know the request.js lib will go looking there for it?

lib/webResource.ts
```diff
@@ -289,9 +305,8 @@ export interface RequestPrepareOptions {
  * The Parameter value provided for path or query parameters in RequestPrepareOptions
  */
 export interface ParameterValue {
-  value: any;
+  value: string;
```
amarzavery: same as above.

lib/webResource.ts
RikkiGibson: Assuming we deal with things missing a `value` property by just calling `.toString()` on them (maybe implicitly through a template string), I think I'd be happiest with `{ toString(): string }` as the type. This allows `number`, `boolean`, `object`, array, etc.

`array` is not a type so it would have to be either any[] or something recursive.

lib/webResource.ts
amarzavery: yup. And I also added a request filter over here https://github.com/Azure/azure-sdk-for-node/pull/2366/files

RikkiGibson: OK, will sort this out before merge

amarzavery: This is the issue that I created https://github.com/Azure/ms-rest-js/issues/8. I also provided links to similat issues that other people reported in those repos. This will be helpful. To provide context: I had to make this change in ms-rest for node sdk because cognitive services needed this.

lib/webResource.ts
RikkiGibson: See discussion https://github.com/Azure/ms-rest-js/pull/21#discussion_r178665476

lib/webResource.ts
amarzavery: I think this is a very big issue on it's own right. For now you can probably revert the changes related to formData and solve them as a part of solving #8.

RikkiGibson: Pushing the formData commit to a different branch to handle later.

lib/webResource.ts
RikkiGibson: Talked it over a bit more and concluded that `any | ParameterValue` makes the most sense for now

PR #33, opened 2018-04-06
Add initial implementation of new httpPipeline

daschult:
This pull request is just an opportunity to provide initial feedback and to see the type of work that I'm doing on the new httpPipeline architecture. Notice that this PR is going from `daschult/httpPipelineDev` to `daschult/htttpPipeline`. I'll only merge to `master` once all of the work is completed and has been merged into `daschult/httpPipeline`.
This PR just adds the new httpPipeline structure without the `OperationSpec` type and without modifying the existing pipeline. This implementation is basically just a general purpose HTTP Pipeline.


lib/httpClientToRequestPolicyAdapter.ts
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+import { HttpClient } from "./httpClient";
+import { HttpRequest } from "./httpRequest";
+import { HttpResponse } from "./httpResponse";
+import { RequestPolicy } from "./requestPolicy";
+
+/**
+ * An adapter type that adapts a HttpClient to look like a RequestPolicy.
+ */
+export class HttpClientToRequestPolicyAdapter implements RequestPolicy {
```
RikkiGibson: Do adapters still need to be a thing when we have structural typing? that is, it seems like HttpClient should be assignable to RequestPolicy in this language

RikkiGibson: I guess RequestPolicy has a named method on it, so not quite. This could maybe be accomplished with something like:

```js
{
    send: httpClient
}
```
but this is bikeshedding a bit

lib/fetchHttpClient.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns {fetch} fetch - The fetch() method available in the environment to make requests
+ */
+function getFetch(): FetchMethod {
+    return require("fetch-ponyfill")({ useCookie: true }).fetch;
+}
+
+/**
+ * The cached fetch method that will be used to send HTTP requests.
+ */
+let fetch: FetchMethod;
```
RikkiGibson: It seems simpler to be eager here.

lib/httpResponse.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Get the body of this HttpResponse as a string.
+     */
+    bodyAsText(): Promise<string | undefined>;
+
+    /**
+     * Get the body of this HttpResponse as a JSON object.
+     */
+    bodyAsJson(): Promise<{} | any[] | undefined>;
```
RikkiGibson: This is the return value of JSON.parse or xml2js.parseString, right? Let's call it parsedBody and IMO to avoid overengineering it just let it be a Promise<any> (since otherwise describing the types of values contained in it requires a recursive type definition that I can't think of a name for)

lib/policies/userAgentRequestPolicyFactory.ts
```diff
@@ -0,0 +1,32 @@
...
+    };
+}
+
+class UserAgentRequestPolicy extends BaseRequestPolicy {
+    constructor(private _userAgent: string, nextPolicy: RequestPolicy, options: RequestPolicyOptions) {
+        super(nextPolicy, options);
+    }
+
+    send(request: HttpRequest): Promise<HttpResponse> {
+        if (this.shouldLog(HttpPipelineLogLevel.INFO)) {
```
RikkiGibson: what do we do in the browser where we can't set User-Agent? Probably not that important

lib/requestPolicy.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Uses the decorator pattern to add custom behavior when an HTTP request is made.
+ * e.g. add header, user agent, timeout, retry, etc.
+ */
+export interface RequestPolicy {
+    /**
+     * Send the provided HttpRequest request.
+     * @param request The HTTP request to send.
+     * @return A Promise that resolves to the HttpResponse from the targeted server.
+     */
+    send(request: HttpRequest): Promise<HttpResponse>;
```
RikkiGibson: I guess I'd advocate for making this a type alias for a function type, or just making this an invokable member (remove the `send` and leave everything else).

lib/requestPolicy.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * A base class implementation of RequestPolicy.
+ */
+export abstract class BaseRequestPolicy implements RequestPolicy {
+    constructor(private _nextPolicy: RequestPolicy, private _options: RequestPolicyOptions) {
```
RikkiGibson: I'd lean toward `public readonly nextPolicy`--getters and setters that encapsulate a field are not really as idiomatic in js

test/fetchHttpClientTests.ts
```diff
@@ -0,0 +1,83 @@
...
+import { fetchHttpClient } from "../lib/fetchHttpClient";
+import { HttpRequest } from "../lib/httpRequest";
+import { HttpResponse } from "../lib/httpResponse";
+
+describe("fetchHttpClient", () => {
+    it("should send HTTP requests", () => {
+        const request = new HttpRequest("GET", "http://www.example.com", {});
+        return fetchHttpClient(request)
+            .then((response: HttpResponse) => {
+                assert.deepStrictEqual(response.request, request);
```
RikkiGibson: Do you have a preference toward asserts over `(2+2).should.equal(4)` or `expect(2+2).to.equal(4)` syntax?

test/httpPipelineLogLevelTests.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { HttpPipelineLogLevel, httpPipelineLogLevelToString } from "../lib/httpPipelineLogLevel";
+import * as assert from "assert";
+
+describe("httpPipelineLogLevelToString()", () => {
+    it(`should work with ERROR`, () => {
+        assert.strictEqual(httpPipelineLogLevelToString(HttpPipelineLogLevel.ERROR), "ERROR");
```
RikkiGibson: Might be worth considering [string enums](https://stackoverflow.com/questions/15490560/create-an-enum-with-string-values-in-typescript#15491832) (the enum value implicitly converts to string) instead of having a conversion method

test/httpPipelineTests.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+import * as assert from "assert";
+import { HttpClient } from "../lib/httpClient";
+import { HttpPipeline } from "../lib/httpPipeline";
+import { HttpRequest } from "../lib/httpRequest";
+import { HttpResponse } from "../lib/httpResponse";
+import { userAgentRequestPolicyFactory } from "../lib/policies/userAgentRequestPolicyFactory";
```
RikkiGibson: IMO let the impls of RequestPolicyFactory just be named e.g. userAgentPolicyFactory instead of userAgentRequestPolicyFactory. This is what we did in Java

test/httpPipelineTests.ts
```diff
@@ -0,0 +1,89 @@
...
+    it("should send requests when no request policies are assigned", () => {
+        const httpClient: HttpClient = (request: HttpRequest) => {
+            return Promise.resolve(new InMemoryHttpResponse(request, 200, {}, "hello"));
+        };
+
+        const httpPipeline = new HttpPipeline([], { httpClient: httpClient });
+
+        const httpRequest = new HttpRequest("GET", "http://www.example.com", {});
+        return httpPipeline.send(httpRequest)
+            .then((response: HttpResponse) => {
```
RikkiGibson: Let's make this an `await` and add the `async` modifier to the test lambda on line 13

RikkiGibson: same for every .then() in this changeset really

typings/lib/httpPipelineLogLevel.d.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * The different levels of logs that can be used with the HttpPipelineLogger.
+ */
+export declare enum HttpPipelineLogLevel {
+    /**
+     * A log level that indicates that no logs will be logged.
+     */
+    OFF = 0,
```
RikkiGibson: Already mentioned in another comment, but consider just assigning string values instead of numeric values here.

lib/httpClientToRequestPolicyAdapter.ts
daschult: I changed HttpClient to be a function type instead of an interface. It makes testing easier and I'm hoping it will make it simpler to create new HttpClients (for the handful of people who want to do that). However, RequestPolicy is still an interface so there needs to be an adapter. I could consider changing RequestPolicy to also be a function type, in which case an adapter wouldn't be necessary...

lib/policies/userAgentRequestPolicyFactory.ts
daschult: That's kind of outside of the scope of the request policy. That would be down in the HttpClient function where it would have to strip away the UserAgent header. What happens if you try to send a request with a UserAgent header in the browser? Does it throw an error, or does it just replace it for you?

RikkiGibson: I think it ignores the value you specified and uses its own.

lib/requestPolicy.ts
daschult: Yeah, I'll change this to be an invokable member. That'll be better for BaseRequestPolicy.

lib/policies/userAgentRequestPolicyFactory.ts
daschult: Then I don't think we need to have extra logic for the browser then. If this essentially becomes a no-op, then I'm fine with leaving it.

test/fetchHttpClientTests.ts
daschult: I definitely prefer assertions. I know it's commonplace in Javascript, but I'm not a fan on libraries injecting stuff into my types/objects.

test/httpPipelineLogLevelTests.ts
daschult: I didn't know that typescript added string enums. I'll change HttpMethod and HttpPipelineLogLevel to use those instead.

test/httpPipelineTests.ts
daschult: Okay.

test/httpPipelineTests.ts
daschult: I'll change it to async/await.

typings/lib/httpPipelineLogLevel.d.ts
daschult: I'll have to make sure that the comparison in RequestPolicyOptions.shouldLog() still works after changing this to a string enum.

lib/httpClientToRequestPolicyAdapter.ts
daschult: Turns out if I change RequestPolicy to have an invokable method on it, then BaseRequestPolicy can't be abstract. It would have to implement the invokable method. That seems to be more trouble than it's worth, so I'm just going to change HttpClient back to being an interface with a send() method on it.

daschult: Now HttpClient implements RequestPolicy, so the adapter goes away. Problem solved.

lib/fetchHttpClient.ts
daschult: Done.

lib/httpResponse.ts
daschult: I'm trying to keep the `bodyAs...()` methods similarly named, so how about we called them `textBody()` and `deserializedBody()`. We don't use the term "parsing" anywhere in our RequestPolicies, but we do we the terms "serializing" and "deserializing", so I feel like "deserialized" would be more familiar and would fit better.

lib/requestPolicy.ts
daschult: I talked about my experience with this in a previous comment. Making this invokable causes problems in BaseRequestPolicy (invokable functions can't be marked as abstract in an abstract base class). I'm going to leave this as a named function.

lib/httpResponse.ts
RikkiGibson: Now I remember that this is about "deserializing in the pipeline". So I understand that this isn't the return value of a JSON/XML.parse method, it's actually the return value of serializer.deserialize. deserializedBody seems fine.

lib/httpClientToRequestPolicyAdapter.ts
RikkiGibson: Given that, still worth considering using structural typing at the usage site of this adapter https://github.com/Azure/ms-rest-js/pull/33#discussion_r179842294 instead of declaring a named adapter class.

That is, `{ send: httpClient }` instead of `new HttpClientToRequestPolicyAdapter(httpClient)`

typings/lib/httpPipelineLogLevel.d.ts
daschult: Changing this to a string enum did break the comparison that I need. I found [this trick](https://stackoverflow.com/a/42172322) though, and it works great.

RikkiGibson: Oh, this is about the enum values having an associated severity, right?

amarzavery:
@daschult - Is it ok, if I review this PR by Monday, as I need to get some stuff done for EventHubs?

daschult:
@amarzavery That's fine.

lib/requestPolicy.ts
daschult: I've removed property usages from my implementation. I thought about what you said, and I decided that they were just overhead/wrappers that didn't add any benefit.

typings/lib/httpPipelineLogLevel.d.ts
daschult: Correct.

daschult: I did change HttpMethod to a string enum, though, because they don't need to be compared relative to each other.

lib/httpHeaders.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Set a header in this collection with the provided name and value. The name is
+     * case-insensitive.
+     * @param headerName The name of the header to set. This value is case-insensitive.
+     * @param headerValue The value of the header to set.
+     */
+    public set(headerName: string, headerValue: string | number): void {
+        this.headersMap[getHeaderKey(headerName)] = { name: headerName, value: headerValue.toString() };
```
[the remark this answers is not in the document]
RikkiGibson: I think until we zero in on how we handle the headers definition in Swagger we should err away from anything more complex than headerValue.toString(). See [the header object documentation](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#headerObject)

lib/httpHeaders.ts
```diff
@@ -0,0 +1,114 @@
...
+export interface HttpHeader {
+    /**
+     * The name of the header.
+     */
+    name: string;
+
+    /**
+     * The value of the header.
+     */
+    value: string;
```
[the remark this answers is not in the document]
daschult: I feel that saying `value: any` doesn't help consumers of this interface at all. I'd rather make everything a string and let the consumer convert it to a number if they know it will be a number.

lib/httpHeaders.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * A HttpHeaders collection represented as a simple JSON object.
+ */
+export type RawHttpHeaders = { [headerName: string]: string };
```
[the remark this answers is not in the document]
daschult: I disagree. These are strings when they come across the wire. Let's reflect that in our HttpHeaders implementation.

lib/httpPipeline.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * A collection of RequestPolicies that will be applied to a HTTP request before it is sent and will
+ * be applied to a HTTP response when it is received.
+ */
+export class HttpPipeline {
+    private readonly httpClient: HttpClient;
+    private readonly requestPolicyOptions: RequestPolicyOptions;
+
+    constructor(private readonly requestPolicyFactories: RequestPolicyFactory[], private readonly options: HttpPipelineOptions) {
```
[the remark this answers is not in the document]
daschult: Okay, I'll switch them all to start with underscores.

lib/httpResponse.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Get the body of this HttpResponse as a string.
+     */
+    textBody(): Promise<string | undefined>;
+
+    /**
+     * Get the body of this HttpResponse after it has been deserialized.
+     */
+    deserializedBody(): Promise<any>;
```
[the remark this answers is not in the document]
daschult: We talked about this, and we decided that the purpose of making these functions is to make them evaluate lazily.

lib/policies/userAgentPolicy.ts
```diff
@@ -0,0 +1,32 @@
...
+class UserAgentPolicy extends BaseRequestPolicy {
+    constructor(private readonly userAgent: string, nextPolicy: RequestPolicy, options: RequestPolicyOptions) {
+        super(nextPolicy, options);
+    }
+
+    send(request: HttpRequest): Promise<HttpResponse> {
+        if (this.shouldLog(HttpPipelineLogLevel.INFO)) {
+            this.log(HttpPipelineLogLevel.INFO, `Set "User-Agent" header to "${this.userAgent}".`);
+        }
+        request.headers.set("User-Agent", this.userAgent);
```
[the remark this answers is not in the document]
daschult: Rikki and I talked about this, and we determined that even if this does get set on the browser, it will essentially be a no-op because the browser will just overwrite it. If we decide that we shouldn't be doing this, even though it's a no-op, then the place to skip it would be when we are constructing the default HttpPipeline, not here in the RequestPolicy.

package.json
```diff
@@ -50,6 +50,7 @@
     "@types/xml2js": "^0.4.2",
     "mocha": "^3.5.3",
     "should": "5.2.0",
+    "source-map-support": "^0.5.4",
```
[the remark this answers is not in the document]
daschult: Sure, I'll make the change now.

package.json
```diff
@@ -66,10 +67,10 @@
   },
   "scripts": {
     "tsc": "tsc -p tsconfig.json",
-    "test": "npm -s run-script tslint && npm -s run-script unit",
-    "unit": "mocha -t 50000 ./dist/test",
+    "test": "npm run tsc && npm -s run-script tslint && npm -s run-script unit",
+    "unit": "mocha --require source-map-support/register -t 50000 ./dist/test",
```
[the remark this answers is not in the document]
daschult: Done.

daschult:
@amarzavery @RikkiGibson Thanks for the reviews!

PR #34, opened 2018-04-09
Browser tests support

RikkiGibson:
- Fixed URL parsing in browser. Our logic for detecting relative URLs was broken in the browser, where a default base URL of `window.location` is used. Instead, simply use url-parse's optional base URL argument.
- Passes an `'include'` argument to fetch which allows cross-origin cookies in the browser for testing LROs. Not sure if this is better implemented as a RequestPolicy--this will fail on any service which sends back the CORS header `Access-Control-Allow-Origin: *`, as CORS forbids storing cookies for a service with a wildcard in the allow-origin header.


RikkiGibson:
See Azure/ms-rest-js#35

PR #36, opened 2018-04-10
Convert filters to RequestPolicies

daschult:


lib/credentials/apiKeyCredentials.ts
```diff
@@ -75,4 +76,41 @@ export class ApiKeyCredentials implements ServiceClientCredentials {
...
+
+  /**
+   * Signs a request with the values provided in the inHeader and inQuery parameter.
+   *
+   * @param httpRequest The HttpRequest to be signed.
+   * @returns The signed request.
+   */
+  public async signHttpRequest(httpRequest: HttpRequest): Promise<HttpRequest> {
+    if (!httpRequest) {
+      throw new Error(`httpRequest cannot be null or undefined and must be of type "object".`);
```
RikkiGibson: Did this previously contain a `typeof` test? Also, I sorta wish that `strictNullChecks` would be adequate here, as TypeScript is supposed to warn if you ever pass a value that could be null/undefined to this method.

RikkiGibson: I figure that in practice, crossing library boundaries and grabbing values over the wire makes this kind of thing more necessary..

daschult: If we could guarantee that our customers would use TypeScript, then the type system would be enough here and I wouldn't need to check whether httpRequest is defined or not. For some of the request policies I tried to copy and paste the existing filter implementation with minimal changes. This is one of them.

lib/credentials/basicAuthenticationCredentials.ts
```diff
@@ -44,4 +45,17 @@ export class BasicAuthenticationCredentials implements ServiceClientCredentials
...
+
+  /**
+   * Signs a request with the Authentication header.
+   *
+   * @param httpRequest The request to be signed.
+   * @returns The signed request.
+   */
+  public signHttpRequest(httpRequest: HttpRequest): Promise<HttpRequest> {
+    const credentials = `${this.userName}:${this.password}`;
+    const encodedCredentials = `${this.authorizationScheme} ${Buffer.from(credentials).toString("base64")}`;
```
RikkiGibson: Is it necessary to specify a character encoding for `Buffer.from`?

lib/httpRequest.ts
```diff
@@ -9,29 +9,22 @@ import { HttpMethod } from "./httpMethod";
...
+    /**
+     * Create a new HTTP request using the provided properties.
+     * @param _httpMethod The HTTP method that will be used to send this request.
+     * @param url The URL that this request will be sent to.
+     * @param headers The HTTP headers to include in this request.
+     * @param _body The body of this HTTP request.
+     */
+    constructor(public httpMethod: HttpMethod, public url: string, headers: HttpHeaders | RawHttpHeaders, private _body?: string) {
+        if (!this.url) {
+            const urlString: string = (this.url === undefined || this.url === null ? this.url : `"${this.url}"`);
```
RikkiGibson: prefer `== undefined`?

lib/credentials/basicAuthenticationCredentials.ts
daschult: I'm not sure. This was one of the implementations that I copied and pasted. @amarzavery Do you know?

lib/credentials/serviceClientCredentials.ts
```diff
@@ -11,4 +12,12 @@ export interface ServiceClientCredentials {
...
    */
   signRequest(webResource: WebResource): Promise<WebResource>;
+
+  /**
+   * Signs a request with the Authentication header.
+   *
+   * @param request The HTTP request to be signed.
+   * @returns The signed request;
+   */
+  signHttpRequest(httpRequest: HttpRequest): Promise<HttpRequest>;
```
amarzavery: TypeScript and Javascript rely on duck typing. As long as the shape remains the same then it should make no difference to consumers. Eventually WebResource will be replaced by HttpRequest. So why have two methods over here? Wont it confuse customers of the node sdk (using ms-rest) today? The runtime is being used a lot by folks inside Microsoft (for developing vscode extensions, etc.) and outside customers as well.

lib/httpRequest.ts
```diff
@@ -9,29 +9,22 @@ import { HttpMethod } from "./httpMethod";
 export class HttpRequest {
     private readonly _headers: HttpHeaders;
 
-    constructor(private _httpMethod: HttpMethod, private _url: string, headers: HttpHeaders | RawHttpHeaders, private _body?: string) {
-        if (!this._url) {
-            const urlString: string = (this._url === undefined || this._url === null ? this._url : `"${this._url}"`);
+    /**
+     * Create a new HTTP request using the provided properties.
+     * @param _httpMethod The HTTP method that will be used to send this request.
```
amarzavery: This should be httpMethod (remove underscore) as it is not private anymore

lib/policies/exponentialRetryPolicy.ts
```diff
@@ -0,0 +1,126 @@
...
+        this._maximumRetryDelayInMilliseconds = retryOptions.maximumRetryIntervalInMilliseconds || 90 * 1000;
+        this._delayFunction = retryOptions.delayFunction || utils.delay;
+    }
+
+    /**
+     * Get whether or not we should retry the request based on the provided response.
+     * @param response The response to read to determine whether or not we should retry.
+     */
+    protected shouldRetry(details: { response?: HttpResponse, responseError?: RetryError}): boolean {
+        let result = true;
```
amarzavery: General comment: make sure the indentation is 2 spaces. That is the norm in JS/TS world.

lib/policies/msRestNodeJsUserAgentPolicy.ts
```diff
@@ -0,0 +1,52 @@
...
+        return new MsRestNodeJsUserAgentPolicy(userAgentInfo, nextPolicy, options);
+    };
+}
+
+class MsRestNodeJsUserAgentPolicy extends BaseRequestPolicy {
+    constructor(private readonly _userAgentInfo: string[], nextPolicy: RequestPolicy, options: RequestPolicyOptions) {
+        super(nextPolicy, options);
+    }
+
+    send(request: HttpRequest): Promise<HttpResponse> {
```
amarzavery: Are we making sure that this filter will only be added or executed for node.js and not the browser in some other way?

test/policies/exponentialRetryPolicyTests.ts
```diff
@@ -0,0 +1,67 @@
...
+import { HttpMethod } from "../../lib/httpMethod";
+import { HttpRequest } from "../../lib/httpRequest";
+import { HttpResponse } from "../../lib/httpResponse";
+import { InMemoryHttpResponse } from "../../lib/inMemoryHttpResponse";
+import { exponentialRetryPolicy } from "../../lib/policies/exponentialRetryPolicy";
+import { RequestPolicy } from "../../lib/requestPolicy";
+import { RequestPolicyFactory } from "../../lib/requestPolicyFactory";
+import { RequestPolicyOptions } from "../../lib/requestPolicyOptions";
+
+describe("exponentialRetryPolicy", () => {
```
amarzavery: General comment: It may be fine for now, but we also need to revisit mechanism to record tests. We have been using a super old version of nock.js for test recording. Now since we are rewriting stuff we should take a fresh stab at it.

lib/policies/msRestNodeJsUserAgentPolicy.ts
RikkiGibson: I'm tempted to go "laissez faire" here--if a browser user tries to do it, we do whatever we do and accept that it won't work in the browser. Otherwise you add a check for the "isNode" flag, flag, someone shows up with a new javascript runtime environment (e.g. react native?) where you can set the user-agent and now we're failing unnecessarily.

lib/credentials/basicAuthenticationCredentials.ts
RikkiGibson: I'm not referring to the "base64" argument to toString--I'm referring to buffer.from. The correct translation of a string into essentially a byte array depends on getting the character encoding right.

lib/credentials/basicAuthenticationCredentials.ts
```diff
@@ -44,4 +45,17 @@ export class BasicAuthenticationCredentials implements ServiceClientCredentials
...
     return Promise.resolve(webResource);
   }
+
+  /**
+   * Signs a request with the Authentication header.
+   *
+   * @param httpRequest The request to be signed.
+   * @returns The signed request.
+   */
+  public signHttpRequest(httpRequest: HttpRequest): Promise<HttpRequest> {
```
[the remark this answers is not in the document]
daschult: This is a temporary change that will be reverted when I change WebResource to HttpRequest. I could've done method overloading here, but rather than learn how to do that in TypeScript, I figured I'd just use a different name for now and then switch it back to signRequest(HttpRequest) when I change to using HttpRequest instead of WebResource.
Just to be clear, changing to use the new HttpPipeline will break customers. This is going to be a breaking change. I understand that it's a good idea to minimize the number of breaking changes, but it's not completely avoidable.

lib/credentials/serviceClientCredentials.ts
daschult: See my previous comment.

lib/httpRequest.ts
daschult: Done.

lib/policies/exponentialRetryPolicy.ts
daschult: Done.

lib/credentials/basicAuthenticationCredentials.ts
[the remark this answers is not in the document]
amarzavery: I see. Sounds good to me. It is always good to avoid breaking changes wherever possible, unless they are absolutely required.

lib/policies/msRestNodeJsUserAgentPolicy.ts
daschult: If a user wants to add this policy to their pipeline, then that's their choice. When we are constructing the default pipeline (hasn't been coded up yet), we'll then check once whether isNode() or not. If we're running in Node, then we'll add the Node-specific functionality, such as adding this policy. If not, then we'll leave it out.

amarzavery: Do we plan to extract common functionality into a base userAgentPolicy and then have node and browser specific policies for user agent?
The name of the policy (msRestNodeJsUserAgentPolicy.ts) indicates that this is intended to be used for node.js. Then why not add the safety check `isNode()` in the definition itself?

amarzavery: In this way the customer will not make errors as well.

daschult: It would probably be a good idea to extract the common functionality into the existing userAgentPolicy type. I didn't do it in this PR, but I can in a future one.
I didn't want to add the isNode() safety check because I don't want it to be executed for every request. I'd rather have some sort of error occur because a customer configured their pipeline incorrectly, rather than silently do extra work.

PR #37, opened 2018-04-10
No build products in git

### LINKED ISSUE #35: Remove build products from git
Let's remove all *.js, *.js.map, and *.d.ts files from version control, as we did in autorest.typescript. This will make our diffs cleaner and make it easier to keep the repo in a consistent state. Conversely, let's use .npmignore to ensure that we don't publish any of our source *.js or *.js.map files in the npm package.

This task also applies to Azure/ms-rest-azure-js.

@amarzavery @daschult I'm thinking that we should have a better place to publish our bundle files than the master branch of the repo. Any site which simply loads a bundle built from the tip of master every time their page loads is bound to break down the line.

How about we use GitHub Releases to make each version of the bundle available via HTTPS? Then users will specify in the URL which version of the library they're trying to load.

RikkiGibson:
Fixes #35. See the discussion there about the question of where to put msRestBundle.js.

- Uses `npm prepare` step so that we run a build before packaging. I've ensured that the output of `npm pack` works with autorest.typescript, so I'm confident that we can publish like this. 
  - This even works when we add ms-rest-js's git repo to a user's package.json!
- Uses .npmignore so that we don't put *.ts sources in the npm package
- Uses npm-run-all to run some build/test steps in parallel
- Uses mocha.opts so that we can just `mocha` in the repo root to run nodejs unit tests-- now passing extra arguments to mocha is super simple
- Removes redundant flags from tsconfig.json that are already covered by `--strict`

Will create a similar PR for ms-rest-azure-js.


.npmignore
```diff
@@ -1,8 +1,13 @@
 .vscode/
-node_modules/
```
daschult: We still want to ignore node_modules, right?

gulpfile.js
```diff
@@ -1,30 +0,0 @@
-var gulp = require("gulp"),
```
daschult: I don't use this gulpfile, but did you remove it on purpose?

daschult: Not used. Cool

amarzavery: This was added by a team mate who made some contributions to ms-rest-js. He found it useful hence added it. I don't think we are using it. Safe to remove.

RikkiGibson: That's good to know. If we for some reason need it back in the future let's make sure to also add gulp as a devDependency. For now, npm-run-all seems great as it lets us run our node and browser builds and tests in parallel with very little fuss.

amarzavery:
General question:
- Would we autogenerate the js and d.ts files before publishing to npm? OR
- Would we publish the typescript source code as-is and then when someone executes `npm install ms-rest-js` then in the pre installation step, npm runs tsc to generate the js and d.ts files and then installs the package?

daschult:
@amarzavery I vote that we generate the js and d.ts files (aka build) before publishing to NPM.

RikkiGibson:
For publishing to npm, we run `npm prepare` which will build everything. The *.js and *.d.ts go into our npm package, and no *.ts sources go into the package. You can verify this by running `npm pack` in this repo.

For cases like adding a local folder path or a git repo/branch to package.json dependencies, `npm prepare` (i.e. build typescript and webpack) will be run on the user's machine automatically.

I hope that clears things up. Looks like this is actually a pretty new feature in npm 5 and it seems to make this kind of use case possible/convenient. :+1:

RikkiGibson:
Actually could stand to be a little more strict with the .npmignore and avoid putting anything in `/dist/test` into the package-- but there's no rush to make this correction.

PR #44, opened 2018-04-16
Add TypeSpecs and tests

daschult:


lib/serialization/base64UrlSpec.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+import { TypeSpec, createValidationErrorMessage } from "./typeSpec";
+import { SerializationOptions } from "./serializationOptions";
+import { SpecPath } from "./specPath";
+
+/**
+ * A type specification that describes how to validate and serialize a Base64Url encoded ByteArray.
+ */
+const base64UrlSpec: TypeSpec<string, Buffer> = {
```
RikkiGibson: Buffer presents a problem in the browser. Right now we have webpack emit basically a 50kb polyfill for it using the "buffer" package. The closest supertype available in the browser is Uint8Array. This PR might not be the time yet, but let's have a task for solving the Buffer problem.

RikkiGibson: This SO post has some decent suggestions about base64 conversions to/from Uint8Array, and some links to MDN that might provide useful context. https://stackoverflow.com/questions/12710001/how-to-convert-uint8-array-to-base64-encoded-string#12713326

lib/serialization/base64UrlSpec.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (!anyValue || !anyValue.constructor || typeof anyValue.constructor.isBuffer !== "function" || !anyValue.constructor.isBuffer(value)) {
+      if (options && options.serializationStrictTypeChecking) {
+        throw new Error(createValidationErrorMessage(propertyPath, value, "a Buffer"));
+      }
+      result = anyValue;
+    } else {
+      const base64String: string = value.toString("base64");
+
+      let trimmedResultLength = base64String.length;
+      while ((trimmedResultLength - 1) >= 0 && base64String[trimmedResultLength - 1] === "=") {
```
RikkiGibson: Was this just migrated from the existing serializer or is it new?

RikkiGibson: At a glance, it seems like we should not have to do something like trim the trailing `=` characters from a base64 encoded string.

lib/serialization/base64UrlSpec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  deserialize(propertyPath: SpecPath, value: string, options: SerializationOptions): Buffer {
+    let result: Buffer;
+
+    if (!value || typeof value !== "string") {
+      if (options && options.deserializationStrictTypeChecking) {
+        throw new Error(createValidationErrorMessage(propertyPath, value, "a string"));
+      }
+
+      result = <any>value;
```
RikkiGibson: From basarat's excellent [TypeScript Deep Dive](https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html):

> ## `as foo` vs. `<foo>`
> Originally the syntax that was added was `<foo>`. This is demonstrated below:
> ```ts
> var foo: any;
> var bar = <string> foo; // bar is now of type "string"
> ```
> However there is an ambiguity in the language grammar when using <foo> style assertions in JSX:
> ```ts
> var foo = <string>bar;
> </string>
>```
> Therefore it is now recommended that you just use `as foo` for consistency.

lib/serialization/booleanSpec.ts
```diff
@@ -0,0 +1,32 @@
...
+import { SerializationOptions } from "./serializationOptions";
+import { SpecPath } from "./specPath";
+
+/**
+ * A type specification that describes how to validate and serialize a boolean.
+ */
+const booleanSpec: TypeSpec<boolean, boolean> = {
+  specType: "boolean",
+
+  serialize(propertyPath: SpecPath, value: boolean, options: SerializationOptions): boolean {
```
RikkiGibson: Are you handling the parsing from string case that xml2js needs somewhere else?

lib/serialization/serializationOptions.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * The type of output that will be produced.
+   */
+  outputType?: SerializationOutputType;
+
+  /**
+   * Whether or not serialization will follow strict type-checking. If strict type-checking is used,
+   * then an Error will be thrown if a value doesn't match the provided TypeSpec's expected types.
+   */
+  serializationStrictTypeChecking?: boolean;
```
RikkiGibson: Is the idea here that Azure devs might opt in to these strictness options based on their personal preference?

lib/serialization/base64UrlSpec.ts
daschult: This was migrated from the existing serializer.
We remove the trailing `=` because this is base64URL, which is slightly different than base64.

RikkiGibson: Cool, makes sense.

test/serialization/base64UrlSpecTests.ts
```diff
@@ -0,0 +1,151 @@
...
+
+describe("base64UrlSpec", () => {
+  it("should have \"Base64Url\" for its typeName property", () => {
+    assert.strictEqual("Base64Url", base64UrlSpec.specType);
+  });
+
+  describe("serialize()", () => {
+    describe("with strict type-checking", () => {
+      function base64UrlSerializeWithStrictTypeCheckingTest(args: { propertyPath?: string[], value: Buffer, expectedResult: string | Error }): void {
+        serializeTest({
```
RikkiGibson: I don't see a definition for this in the diff--is this basically the same suite just migrated based on the new serializer structure?

daschult: Yeah, basically.

lib/serialization/serializationOptions.ts
daschult: Personal preference, or for debugging. I actually just used these options this morning to detect a problem with an operation spec.
@amarzavery mentioned that sometimes resource provider teams will return types that are different than what the swagger indicates they'll be. This shouldn't happen, but sometimes it does. In those cases, it's better to allow the customer to specify that they want the deserialization to be looser so that they can get the response and work with it, even if it doesn't match the specification perfectly.

lib/serialization/base64UrlSpec.ts
daschult: I'll definitely need your guys' input when it comes to the browser, because I've never done anything with browsers.
Let's handle Buffers and Streams in seperate PRs.

lib/serialization/booleanSpec.ts
daschult: I'll do that in the SerializationPolicy.

lib/serialization/specPath.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+export class SpecPath {
```
amarzavery: missing documentation for SpecPath that describes its purpose. Will be helpful for future developers.

daschult: Good point. I'll add that.

lib/serialization/timeSpanSpec.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+import { TypeSpec, createValidationErrorMessage } from "./typeSpec";
+import { isDuration, Duration, duration } from "moment";
```
amarzavery: At some point we want to move away from moment or extract out duration specific parts of moment and create a separate library. We need moment only for duration and moment brings ~60KB along with it in the runtime, thus making the runtime bulky.

daschult: I think that's a great idea. I've created [this issue](https://github.com/Azure/ms-rest-js/issues/47) for tracking that.

PR #49, opened 2018-04-16
Add SerializationPolicy and rename SpecPath to PropertyPath

daschult:


lib/fetchHttpClient.ts
```diff
@@ -34,7 +36,7 @@ export class FetchHttpClient implements HttpClient {
       statusCode: fetchResponse.status,
       headers: responseHeaders,
       textBody: async () => await fetchResponse.text(),
-      deserializedBody: async () => await fetchResponse.json()
+      serializedBody: async () => await fetchResponse.json()
```
RikkiGibson: Note that [json parses the body content](https://developer.mozilla.org/en-US/docs/Web/API/Body/json) so I think the property name doesn't seem to fit with the implementation here.

lib/httpResponse.ts
```diff
@@ -23,13 +23,18 @@ export interface HttpResponse {
...
    * Get the body of this HttpResponse as a string.
    */
   textBody(): Promise<string | undefined>;
 
   /**
-   * Get the body of this HttpResponse after it has been deserialized.
+   * Get the body of this HttpResponse before it has been deserialized.
    */
-  deserializedBody(): Promise<any>;
+  serializedBody(): Promise<any | undefined>;
```
RikkiGibson: I'd be happier with `rawBody` or `body` alongside `deserializedBody` so that we don't imply that the body content is always something deserializable. A response body stream also needs to be exposed, but I think that will happen later.

test/policies/serializationPolicyTests.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    const inMemoryEchoServer: RequestPolicy = {
+      send: (request: HttpRequest) => {
+        return Promise.resolve(new InMemoryHttpResponse(request, 200, request.headers, JSON.stringify(request.serializedBody)));
+      }
+    };
+
+    const policy: RequestPolicy = policyFactory(inMemoryEchoServer, new RequestPolicyOptions());
+
+    const request = new HttpRequest(
```
RikkiGibson: Consider modifying this constructor to take an object (allows for labeled arguments by convention, and passing a subset of optional arguments without having to `foo, null, null, bar`). e.g.

```ts
new HttpRequest({
  method: HttpMethod.GET,
  url: ...,
  headers: ...,
  body: ...,
  operationSpec: ...
})
```

daschult: I agree. I'll do that.

lib/httpResponse.ts
daschult: How about calling it `jsonBody()`? It's the equivalent of calling `JSON.parse(await this.textBody())`, so that seems to fit to me.

daschult: Eh... maybe `parsedBody()` would work. Funny, that's what you and Amar suggested at the beginning of this work.

RikkiGibson: If you want to expose basically the result of JSON.parse, which I think the current runtime design does on its HttpOperationResponse type, parsedBody (since it might be parsed XML) is fine.

However since serialization is moving into the pipeline it seems like you might as well expose the response body in just 2 forms:
1. the raw form, probably a [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)
2. the parsed-then-deserialized (by an OperationSpec/TypeSpec) object tree version of the body

daschult: Okay, okay. You're right. There really should only be two ways to get the body of the response:
1. `textBody()`
2. `deserializedBody()`
I've made this change.

test/policies/serializationPolicyTests.ts
daschult: Done.

PR #55, opened 2018-04-17
Change ServiceClient to use HttpPipeline

daschult:


lib/httpPipeline.ts
```diff
@@ -124,6 +124,10 @@ export class HttpPipeline {
    * @return A Promise that resolves to the HttpResponse from the targeted server.
    */
   public send(request: HttpRequest): Promise<HttpResponse> {
+    if (!request || !(request instanceof HttpRequest)) {
```
RikkiGibson: Just `!(request instanceof HttpRequest)` will do-- falsy values will return false for `instanceof HttpRequest`

lib/serialization/dictionarySpec.ts
```diff
@@ -19,7 +19,7 @@ export interface DictionaryTypeSpec<TSerializedValue, TDeserializedValue> extend
 /**
  * A type specification that describes how to validate and serialize a Dictionary of values.
  */
-export default function dictionarySpec<TSerializedValue, TDeserializedValue>(valueSpec: TypeSpec<TSerializedValue, TDeserializedValue>): DictionaryTypeSpec<TSerializedValue, TDeserializedValue> {
+export function dictionarySpec<TSerializedValue, TDeserializedValue>(valueSpec: TypeSpec<TSerializedValue, TDeserializedValue>): DictionaryTypeSpec<TSerializedValue, TDeserializedValue> {
```
RikkiGibson: Did you run into weirdness/inconsistency with this, or just have other stuff to export?

test/data/TestClient/lib/testClient.ts
```diff
@@ -12,10 +12,10 @@
 /* jshint forin:false */
 /* jshint noempty:false */
 
-'use strict';
+"use strict";
```
RikkiGibson: `--alwaysStrict` which is implied by `--strict` will always put "use strict" in the output .js file. No need to go on a hunt to remove them all though.

PR #57, opened 2018-04-17
Browser tests

RikkiGibson:
Hopefully will be of use.

For now no CI step--you should `npm run build:browser && npm run test:chrome-unit` in the simple case. It's not that fast, but maybe will refine it later.

### DIFF
--- a/test/userAgentFilterTests.ts
+++ b/test/userAgentFilterTests.ts
@@ -5,13 +5,17 @@ import * as assert from "assert";
 import { WebResource } from "../lib/webResource";
 import { MsRestUserAgentFilter } from "../lib/filters/msRestUserAgentFilter";
 import { Constants } from "../lib/util/constants";
+import { isNode } from "../lib/msRest";
 
 const should = require("should");
 const userAgentHeader = Constants.HeaderConstants.USER_AGENT;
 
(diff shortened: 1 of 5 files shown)


test/userAgentFilterTests.ts
```diff
@@ -5,13 +5,17 @@ import * as assert from "assert";
 import { WebResource } from "../lib/webResource";
 import { MsRestUserAgentFilter } from "../lib/filters/msRestUserAgentFilter";
 import { Constants } from "../lib/util/constants";
+import { isNode } from "../lib/msRest";
```
daschult: I've been trying to not import from `msRest`, but I'm not sure if it's important. The `isNode` variable lives inside a file in the `utils` folder.

RikkiGibson: Oh, I see how it's kinda weird to use the "external" interface when we have access to the "internal" one.

RikkiGibson: I think there's weird stuff going on here related to webpack `externals` config and msRest. I would like to be able to test the actual msRestBundle.js that's distributed, but basically imports that reach into the internal paths seem to result in redundant copies of stuff being included in the test bundle. Will look at it later down the line.

PR #58, opened 2018-04-17
Add TypeSpec name reference support to Sequence and Dictionary TypeSpecs

daschult:


lib/serialization/dictionarySpec.ts
```diff
@@ -60,9 +72,21 @@ export function dictionarySpec<TSerializedValue, TDeserializedValue>(valueSpec:
...
+        if (typeof valueSpec === "string") {
+          if (!options.compositeSpecDictionary || !options.compositeSpecDictionary[valueSpec]) {
+            const errorMessage = `Missing composite specification entry in composite type dictionary for type named "${valueSpec}" at property ${propertyPath}.`;
+            log(options, HttpPipelineLogLevel.ERROR, errorMessage);
+            throw new Error(errorMessage);
+          }
+          valueTypeSpec = options.compositeSpecDictionary[valueSpec] as TypeSpec<TSerializedValue, TDeserializedValue>;
+        } else {
+          valueTypeSpec = valueSpec;
+        }
```
amarzavery: Since serialize and deserialize have the same code repeated twice for dictionary and for sequence and I am sure it would be the same for a CompositeType referencing another CompositeType, how about making this a function and then calling that function in different places. This avoids code duplication and reduces copy paste errors.

amarzavery:
do write a test case for this scenario.

daschult:
@amarzavery I refactored the composite TypeSpec resolution and added tests in https://github.com/Azure/ms-rest-js/pull/58/commits/508016b4c351369abeb017c62378e7be48d6e73a.

amarzavery:
NOTE: Whatever we are doing right now is the standard way of handling errors. This comment is just an FYI about what node.js has in store for us in the near future.

Some good breaking changes are coming for Error objects in node.js 10 which is supposed to be released by end of this month and will go to LTS in Fall 2018. You can find more about it over [here](https://medium.com/the-node-js-collection/node-js-errors-changes-you-need-to-know-about-dc8c82417f65).

lib/serialization/serializationOptions.ts
```diff
@@ -76,6 +77,47 @@ export enum SerializationOutputType {
   XML
 }
 
+export function failSerializeTypeCheck(options: SerializationOptions, propertyPath: PropertyPath, value: any, expectedTypeDescription: string): void {
```
amarzavery: When we say that the type of a parameter is `any` that means all bets are off w.r.t it's type. It could be a string, boolean, object, etc. or it could be null or undefined. The latter part (null or undefined) makes it optional.

In the above signature type of value is `any` which gives it the notion of it being optional. However, the next parameter after it `expectedTypeDescription: string` is a required parameter. This is an anti-pattern with defining a function where all required parameters should come before optional parameters.

It would be nice to flip the order of last two parameters as follows:
```js
export function failSerializeTypeCheck(options: SerializationOptions, propertyPath: PropertyPath, expectedTypeDescription: string, value: any): void;
```
This makes it easier for  the caller of this method to not provide value if it is not present.

amarzavery: On another thought, I dont expect these methods to be called by customers. So we should be fine in this situation. However we should make sure we don't do this for a public facing API.

lib/serialization/compositeSpec.ts
```diff
@@ -83,14 +66,10 @@ export function compositeSpec(typeName: string, propertySpecs: { [propertyName:
               // XML
               if (childPropertySpec.xmlIsWrapped) {
                 if (!childPropertySpec.xmlName) {
-                  const errorMessage = `When the serialization output type is XML, property specification for ${propertyPath.pathStringConcat(childPropertyName)} specified xmlIsWrapped but doesn't have an xmlName value.`;
-                  log(options, HttpPipelineLogLevel.ERROR, errorMessage);
-                  throw new Error(errorMessage);
+                  throw logAndCreateError(options, `When the serialization output type is XML, property specification for ${propertyPath.pathStringConcat(childPropertyName)} specified xmlIsWrapped but doesn't have an xmlName value.`);
```
RikkiGibson: good--meant to suggest something like this earlier.

lib/serialization/serializationOptions.ts
daschult: `value` is definitely not optional. `value` could be templatized, but I thought it was pointless because the type wasn't used anywhere. I understand the anti-pattern of optional parameters before required parameters, but does knowing `value` is required change your opinion?

amarzavery: value is required but it can still be null or undefined, right?

daschult: Yup, it can. And those are valid values for the `value` parameter.

PR #60, opened 2018-04-20
Add testserver, browser tests and stream tests

RikkiGibson:
- Modifies HttpRequest/Response types to use BodyInit-based request bodies and more flexible response bodies. Still deciding which response body types are worth forwarding-- I looked at the GitHub polyfill and it seems that [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) is pretty much a common denominator that lets you convert to everything-- seems to represent a handle on data that's not necessarily in memory.

- No more calls to external servers in test suites-- I added a simple express app under testserver/ which serves up static content as well as webpack test bundles as needed

- Adds browser tests which were merged into master. To run the browser tests I suggest:
1. Start `npm run test:server` in a separate window, then
2. Either `npm run test:chrome-unit` or open localhost:3000 in browser for rapid re-running/debugging of browser tests

- Adds node and browser subdirectories to contain tests that reference nodejs-only or browser-only types. I'm not done with this. I want to redo the files tests based on this and remove the `Buffer: true` in webpack.config.ts. I thought you might benefit from the current state of this for testing, and merging early and often seems good.


RikkiGibson:
mocha-chrome doesn't really seem to be cutting the mustard--will probably look into karma pretty soon.

lib/inMemoryHttpResponse.ts
```diff
@@ -34,4 +34,13 @@ export class InMemoryHttpResponse implements HttpResponse {
   deserializedBody(): Promise<any | undefined> {
     return Promise.resolve(this._deserializedBody);
   }
-}
\ No newline at end of file
+
+  get readableStreamBody(): ReadableStream | NodeJS.ReadableStream {
```
daschult: Don't forget the `null` type in the return type of this method.

RikkiGibson:
Also, I tried to use the paths/baseUrl option in tsconfig.json to make it so we don't need to import from `"../../../../../../lib/whatever"` but the solutions that make it work at runtime in nodejs don't seem mature enough for me to be comfortable depending on them.

amarzavery:
:shipit:

PR #63, opened 2018-04-20
Add support for polymorphism in compositeSpec.ts

daschult:
I highly recommend looking at animal, cat, tiger, and saberToothedTiger in [compositeSpecTests.ts](https://github.com/Azure/ms-rest-js/compare/daschult/httpPipeline...daschult/polymorphism?expand=1#diff-756e21876e0f16d012aeb6b8b34c433cR94) for a good starting point into how polymorphism looks with TypeSpecs.


test/shared/serialization/compositeSpecTests.ts
```diff
@@ -213,63 +332,309 @@ describe("compositeSpec", () => {
...
+        }),
         value: {
           "A": "B doesn't exist in the composite TypeSpec dictionary"
         },
         expectedResult: new Error(`Missing composite specification entry in composite type dictionary for type named "B" at a.property.path.A.`),
         expectedLogs: [`ERROR: Missing composite specification entry in composite type dictionary for type named "B" at a.property.path.A.`]
       });
+
+      compositeSerializeWithStrictTypeCheckingTest({
+        testName: "should log and throw an error when a the composite spec is base type and value is missing property in base type",
```
RikkiGibson: "when the composite spec..."

lib/serialization/compositeSpec.ts
```diff
@@ -8,29 +8,94 @@ export interface CompositeType {
...
+}
+
+/**
+ * The options that specify polymorphism for a CompositeTypeSpec.
+ */
+export interface Polymorphism {
+  /**
+   * The base types for this CompositeTypeSpec.
+   */
+  inheritsFrom?: (CompositeTypeSpec | string) | (CompositeTypeSpec | string)[];
```
amarzavery: The parentheses make it a little confusing while viewing. In reality it could just be oneOf the 4 combinations. So how about doing something like this:
```js
CompositeTypeSpec | string | CompositeTypeSpec[] | string[]
```

RikkiGibson: I think `(CompositeTypeSpec | string)[]` is a different type than `CompositeTypeSpec[] | string[]`. But I might not fully understand how `|` distributes in TypeScript.

RikkiGibson: I actually think removing the first set of parens would help readability.

lib/serialization/compositeSpec.ts
```diff
@@ -8,29 +8,94 @@ export interface CompositeType {
   [key: string]: any;
 }
 
+export interface PolymorphicDerivedType {
+  /**
+   * A CompositeTypeSpec that "inherits" from this CompositeTypeSpec.
+   */
```
amarzavery: Haven't taken a look at other changes yet. However, I hope there is a DerivedType spec that is not polymorphic but is a simple allOf relationship, for ex: StorageAccount is an allOf Resource.

lib/serialization/compositeSpec.ts
```diff
@@ -8,29 +8,94 @@ export interface CompositeType {
...
    */
-  propertySpecs: { [propertyName: string]: PropertySpec };
+  propertySpecs?: { [propertyName: string]: PropertySpec };
 }
 
 /**
  * A type specification that describes how to validate and serialize a Composite value.
  */
-export function compositeSpec(typeName: string, propertySpecs: { [propertyName: string]: PropertySpec }): CompositeTypeSpec {
+export function compositeSpec(args: { typeName: string, polymorphism?: Polymorphism, propertySpecs?: { [propertyName: string]: PropertySpec } }): CompositeTypeSpec {
```
amarzavery: Rather than defining the type inline, do you think defining an interface called CompositeSpecParameters would be better. That makes things cleaner and provides a better intellisense experience as someone can F12 on the type and find out its properties.

amarzavery: nit: Instead of using args can you use parameters. Initially i got confused with args and javascript reserved word `arguments` for getting the arguments array of a function.

lib/serialization/compositeSpec.ts
```diff
@@ -129,76 +201,188 @@ export function compositeSpec(typeName: string, propertySpecs: { [propertyName:
...
+function deserializationDiscriminatorPropertyNameGetter(derivedTypes: PolymorphicDerivedTypes): string {
+  return derivedTypes.discriminatorPropertySerializedName || derivedTypes.discriminatorPropertyName;
+}
+
+/**
+ * Get all of the CompositeTypeSpecs that the provided value must implement if the starting
+ * CompositeTypeSpec is the provided compositeSpec.
+ */
+function getAllPolymorphicCompositeTypeSpecs(value: CompositeType, propertyPath: PropertyPath, compositeSpec: CompositeTypeSpec, options: SerializationOptions, discriminatorPropertyNameGetter: (derivedTypes: PolymorphicDerivedTypes) => string): CompositeTypeSpec[] {
+  const result: CompositeTypeSpec[] = [];
```
amarzavery: What happens if we cannot find a polymorphicCompositeType for a given discriminator value? Do we then serialize or deserialize as that type or throw an Error saying that we could not find a compositeType?

Make sure it is the former as we added support for that in ms-rest for cognitive services.

lib/serialization/compositeSpec.ts
```diff
@@ -129,76 +201,188 @@ export function compositeSpec(typeName: string, propertySpecs: { [propertyName:
...
+
+  let compositeSpecChanged = true;
+  while (compositeSpecChanged) {
```
amarzavery: Is this a one time search while initializing the client or we do this search every time we serialize or deserialize a polymorphic composite type?
 IMO doing it once during initialization and then saving the result as a tree where every node contains an array of  itself and all its children would be a better approach. Then during [de]serialization we simply check if we can find an item in that array. If yes then we are good, otherwise just use the name of the node (i.e that type (which could be the uber base type or an intermediate base type) in the polymorphic tree).

test/shared/serialization/compositeSpecTests.ts
daschult: Fixed in https://github.com/Azure/ms-rest-js/pull/63/commits/f63136568a28c87b9e35249566ad20090fe0064c.

lib/serialization/compositeSpec.ts
daschult: I'll remove the first set of parentheses. @RikkiGibson is correct that the intent is that it can be an array of a combination of CompositeTypeSpecs and strings.

daschult: Fixed in https://github.com/Azure/ms-rest-js/pull/63/commits/f40e92e22aef9ee866814a758f67b0d7773a5c3f.

lib/serialization/compositeSpec.ts
daschult: To my knowledge there is no difference between a `PolymorphicDerivedType` and a `DerivedType`. Since we're talking about TypeSpecs, then a `PolymorphicDerivedType` is just a link from a base-type to a derived-type. All of our polymorphism relationships are allOf relationships.
Can you think of a scenario where this isn't valid?

lib/serialization/compositeSpec.ts
daschult: Fixed in https://github.com/Azure/ms-rest-js/pull/63/commits/89261f3a845e1cd0ad505743b7cfc48428440af7.

lib/serialization/compositeSpec.ts
daschult: I created issue https://github.com/Azure/ms-rest-js/issues/64 for tracking this optimization work.

lib/serialization/compositeSpec.ts
daschult: My current implementation is to throw an error indicating that I couldn't find the requested polymorphic CompositeTypeSpec. I'll make this configurable.

lib/serialization/compositeSpec.ts
amarzavery: how about doing it like this:
```
export type ParentTypeOrParentName = CompositeType | string;

inheritsFrom?: ParentTypeOrParentName | ParentTypeOrParentName[]
```

lib/serialization/compositeSpec.ts
amarzavery: As discussed, an even better approach would be to let the generator generate this information in the model spec as an array of model type names containing the name of that model and all of its children. Then at runtime we do a simple look up in that array. If we can find something then that is awesome. Then we [de]serialize as that child type otherwise we [de]serialize as that (parent) model type and log a warning that we could not find the provided model type.

lib/serialization/compositeSpec.ts
daschult: Done in https://github.com/Azure/ms-rest-js/pull/63/commits/f65d798c55cb4262eb43feb80afbba9a39a6c5fc.

PR #67, opened 2018-04-26
 Revert HttpRequest.method to be only HttpMethod enum, add more exports to msRest

daschult:


lib/httpRequest.ts
```diff
@@ -53,7 +53,7 @@ export class HttpRequest {
   /**
    * The HTTP method of this request.
    */
-  public method: HttpMethod | keyof typeof HttpMethod;
+  public method: HttpMethod;
```
RikkiGibson: This isn't API we expect Azure devs to use very often, but I believe it's not in the spirit of the language to forbid a literal string argument in a situation like this. The string union + namespace of const strings approach would at least clean up the appearance of this property type.

daschult: I did this because I was running into type-system problems when I was trying to read the HttpMethod of a HttpRequest. I'll see if I can find a nicer solution that satisfies both of us.

RikkiGibson: Understandable :)

daschult: I think you'll be happy with the results of [this PR](https://github.com/Azure/ms-rest-js/pull/69).

PR #71, opened 2018-04-27
[Do not merge] Remove Buffer polyfill 

RikkiGibson:
This work is from a little while back. I wanted a PR for it around so I don't lose track of the work. IIRC, it's not quite ready.

This uses the browser Uint8Array and `btoa/atob` APIs (can you think of a worse name for a function?) so that we don't have to include a 50KB~ polyfill for the node.js Buffer type. It changes the byte[] and similar types in the serializer to expect only a Uint8Array, which is a parent type of Buffer.

- [ ] Replace `"is-buffer"` with `instanceof Uint8Array`


RikkiGibson:
This won't get merged as we're not moving forward with the httpPipeline branch

PR #73, opened 2018-05-01
XML serialization in pipeline

RikkiGibson:
This adds coverage of a decent number of cases with XML. I expect there are still more edge cases to handle with XML. I'm kinda leaning toward deferring the remainder of this work until we're at the point where we can start leaning on the autorest.typescript XML tests.

- Changes semantics of SerializationOptions by shallow copying the default options, then overwriting property-by-property with the user options. Can revert if this doesn't seem right.
- Adds XML metadata for root elements to TypeSpec and SequenceSpec
  - This metadata is only meant to be specified when the TypeSpec represents the root of the object tree being serialized. The xml metadata on a PropertySpec should have higher precedence than any metadata specified directly on a TypeSpec.


RikkiGibson:
CI failure:
```sh
ERROR: /home/travis/build/Azure/ms-rest-js/lib/inMemoryHttpResponse.ts[39, 12]: Use 'undefined' instead of 'null'
```
I have to suppress this linter warning or reconfigure the linter to allow a null literal to be returned in a method that doesn't allow returning undefined..

lib/serialization/booleanSpec.ts
```diff
@@ -18,11 +18,19 @@ const booleanSpec: TypeSpec<boolean, boolean> = {
   },
 
   deserialize(propertyPath: PropertyPath, value: boolean, options: SerializationOptions): boolean {
+    if (typeof value === "string" && options.outputType === SerializationOutputType.XML) {
+      if (value === "true") {
```
daschult: I think the type of value should be `boolean | string` now. Since having a string is valid in the XML case, it's a bonus perk that we're more flexible in the JSON scenario. That way you also don't have to do the check for whether we're expecting it to be XML or not?

lib/serialization/numberSpec.ts
```diff
@@ -18,11 +18,18 @@ const numberSpec: TypeSpec<number, number> = {
   },
 
   deserialize(propertyPath: PropertyPath, value: number, options: SerializationOptions): number {
+    if (typeof value === "string" && options.outputType === SerializationOutputType.XML) {
```
daschult: `number | string` here too.

lib/serialization/sequenceSpec.ts
```diff
@@ -1,24 +1,41 @@
...
+}
+
+export interface SequenceSpecOptions {
+  /**
+   * The element name of the root element. Only used when this is a root list.
+   */
+  xmlRootName: string;
+
+  /**
+   * The element name of the sequence elements. Only used when this is a root list.
```
daschult: These comments that say `Only used when this is a root list.` seem to be telling me that they belong on a different kind of `TypeSpec`. I think we should create a new `TypeSpec` named `RootSequenceTypeSpec` that extends `SequenceTypeSpec` and includes these extra properties. That way the extra properties can be required and then they can go at the front of the creator function's (`rootSequenceSpec()`) parameter list.
Does that sound reasonable to you?

lib/serialization/unixTimeSpec.ts
```diff
@@ -23,6 +23,13 @@ const unixTimeSpec: TypeSpec<number, Date> = {
   },
 
   deserialize(propertyPath: PropertyPath, value: number, options: SerializationOptions): Date {
+    if (typeof value === "string" && options.outputType === SerializationOutputType.XML) {
```
daschult: `..., value: number | string, ...`

lib/util/utils.ts
```diff
@@ -229,6 +229,20 @@ const XML2JS_PARSER_OPTS: xml2js.OptionsV2 = {
...
+  return new Promise(function(resolve: (result: any) => void, reject: (err: any) => void) {
+    xmlParser.parseString(xmlString, function(err: any, result: any) {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(result);
+      }
+    });
+  });
+}
```
daschult: Can this function just be async instead of creating and returning a Promise?

test/shared/policies/serializationPolicyTests.ts
```diff
@@ -144,4 +147,91 @@ describe("serializationPolicy", () => {
...
+    };
+
+    const policy: RequestPolicy = policyFactory(inMemoryEchoServer, new RequestPolicyOptions());
+
+    const bodySpec = compositeSpec({
+      typeName: "Root",
+      xmlRootName: "my-root",
+      propertySpecs: {
+        "foo": {
+          xmlName: "foo",
```
daschult: Since the xmlName and the property name are the same (`foo`), xmlName shouldn't be necessary here.

lib/serialization/booleanSpec.ts
RikkiGibson: @amarzavery and I hashed this out in the old serializer. The thinking is that as much as possible we want to give to the user the values that were produced by the service, except in this case with XML, where the document itself is actually ambiguous (whether the string "true" should be interpreted as string or boolean depends on the TypeSpec)

I'm happy to widen the parameter type though.

test/shared/policies/serializationPolicyTests.ts
RikkiGibson: Perhaps needs to be expanded into coverage that handles a default case (xmlName === propertyName) as well as one where the xmlName is different than the property name.

daschult: Sure. Another test where the xmlName isn't the same as the property name sounds good.

lib/serialization/sequenceSpec.ts
daschult: Talked this through. It's fine to just keep this the way it is.

lib/util/utils.ts
RikkiGibson: This is promisifying a callback-only API. Sometime soon, I hope it will simply be synchronous

PR #74, opened 2018-05-02
Bump to v0.3.2. Use isomorphic-xml2js.

RikkiGibson:
This appears to reduce the webpack bundle size from ~245kb -> ~121kb while still passing all the ms-rest-js and autorest.typescript tests.

as an aside, I'm wondering whether it's worth keeping package-lock.json. It seems to create thousands of lines of diff every time it's recreated. Reproducible builds are great--it's just a little frustrating to have to blow this lockfile away so often to get the latest versions of packages. Maybe there's a better way to use npm that doesn't require doing that.

### DIFF
--- a/lib/util/utils.ts
+++ b/lib/util/utils.ts
@@ -7,7 +7,7 @@ import { WebResource } from "../webResource";
 import { Constants } from "./constants";
 import { RestError } from "../restError";
 import { HttpOperationResponse } from "../httpOperationResponse";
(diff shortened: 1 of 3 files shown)


amarzavery:
what version of npm are you using? 6.x ? It seems they made some improvements in 6.x

RikkiGibson:
I'm using 6 and it does seem better than it was before.

PR #75, opened 2018-05-03
Replace momentjs with simple duration format test

RikkiGibson:
This brings us from 121kB -> ~~54kB~~ 68 kB (due to targeting es5).

I tweaked the webpack config so that when we build typescript to target webpack we use es6 modules instead of commonjs. It sounds like this enables more intelligent forms of code splitting and reduced the size by about 2kb. See https://webpack.js.org/guides/tree-shaking/. I think the "es6" module mode and the "es2015" mode are the same thing but the TypeScript docs don't make it crystal clear.


RikkiGibson:
Also this should probably be a bump to 0.4.0

tsconfig.webpack.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "es6",
+    "target": "es2017"
```
amarzavery: do we want to target es2017? Shouldn't this be es5? So we know that stuff works on all the browsers.

RikkiGibson: Yeah, you're right. This should change back to es5, probably by just removing "target" from this file.

lib/util/utils.ts
```diff
@@ -386,4 +386,13 @@ export function applyMixins(targetCtor: any, sourceCtors: any[]): void {
...
     });
   });
-}
\ No newline at end of file
+}
+
+const validateISODuration = /^(-|\+)?P(?:([-+]?[0-9,.]*)Y)?(?:([-+]?[0-9,.]*)M)?(?:([-+]?[0-9,.]*)W)?(?:([-+]?[0-9,.]*)D)?(?:T(?:([-+]?[0-9,.]*)H)?(?:([-+]?[0-9,.]*)M)?(?:([-+]?[0-9,.]*)S)?)?$/;
+
+/**
+ * Indicates whether the given string is in ISO 8601 format.
```
amarzavery: /*
 * Indicates whether the given string is in ISO 8601 format.
 * @param {string} value - The value to be validated for ISO 8601 duration format.
 * @return {boolean} result - `true` if valid, `false` otherwise.
 */
jsdoc comment format, since this will be a public facing method and will end up on docs.microsoft.com

amarzavery:
otherwise LGTM.

XiaoningLiu:
`ES6` and `ES2015` should be same, there is a `/` between them in http://www.typescriptlang.org/docs/handbook/compiler-options.html.

>Specify ECMAScript target version: "ES3" (default), "ES5", "ES6"/"ES2015", "ES2016", "ES2017" or "ESNext".

PR #76, opened 2018-05-04
Change the ServiceClient HttpPipeline to meet storage's immediate needs

daschult:
This transforms the existing before/after pipeline into a RequestPolicy chain pipeline. It does have some breaking changes (ServiceClientOptions.filters -> ServiceClientOptions.requestPolicyCreators), but not many. In this round, I kept the WebResource and HttpOperationResponse the same and only modified the pipeline. All of the autorest.typescript tests pass with this change.


lib/httpPipelineLogger.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Log the provided message.
+   * @param logLevel The HttpLogDetailLevel associated with this message.
+   * @param message The message to log.
+   * @param formattedArguments A variadic list of arguments that should be formatted into the
+   *                           provided message.
+   */
+  log(logLevel: HttpPipelineLogLevel, message: string): void {
+    console.log(`${HttpPipelineLogLevel[logLevel]}: ${message}`);
```
RikkiGibson: This is not a blocker or anything, but consider using the console.warn, console.error etc variants depending on the log level

lib/policies/msRestUserAgentPolicy.ts
```diff
@@ -1,20 +1,27 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-import { BaseFilter } from "./baseFilter";
-import { WebResource } from "../webResource";
+import * as os from "os";
```
RikkiGibson: Any sense of what the footprint is for this in the browser? Consumers can specify `node: { os: "empty" }` in their webpack.config.js but if they forget to do it their bundle will be a bit fatter than strictly necessary.

lib/httpPipelineLogger.ts
daschult: That's a good idea. I'll do that.

lib/policies/rpRegistrationPolicy.ts
```diff
@@ -1,21 +1,32 @@
...
+  };
+}
+
+export class RPRegistrationPolicy extends BaseRequestPolicy {
 
-  constructor(retryTimeout = 30) {
-    super();
+  constructor(nextPolicy: RequestPolicy, options: RequestPolicyOptions, retryTimeout = 30) {
+    super(nextPolicy, options);
     retryTimeout = retryTimeout;
```
RikkiGibson: This isn't part of your work, but this smells funny and should be revisited.

test/userAgentFilterTests.ts
```diff
@@ -3,21 +3,30 @@
...
     if (!isNode) {
       this.skip();
     }
 
     const userAgentArray: Array<string> = [];
-    const userAgentFilter = new MsRestUserAgentFilter(userAgentArray);
+    const userAgentFilter = new MsRestUserAgentPolicy(emptyRequestPolicy, new RequestPolicyOptions(), userAgentArray);
     const resource = new WebResource();
     resource.headers = {};
     userAgentFilter.before(resource).then((resource) => {
```
RikkiGibson: Does this test still make sense if the RequestPolicy is only supposed to have `.sendRequest()` on it?

lib/httpPipelineLogger.ts
daschult: Fixed in https://github.com/Azure/ms-rest-js/pull/76/commits/7d36607f46420aae08b8d7824b13398007bbdcc0.

lib/policies/msRestUserAgentPolicy.ts
daschult: I don't know what the footprint is, but this RequestPolicy should only be added in the Node.js scenario. I've updated the createdDefaultRequestPolicyCreators() method to only add this RequestPolicy in the Node.js scenario.

lib/policies/rpRegistrationPolicy.ts
daschult: Agreed. I'll take care of that right now.

test/userAgentFilterTests.ts
daschult: This test makes sense because it's verifying the behavior of MsRestUserAgentPolicy.before(). I think it makes sense to rename the before() and after() methods that exist on specific RequestPolicies. I can do that now.

lib/policies/rpRegistrationPolicy.ts
daschult: Done.

PR #80, opened 2018-05-08
Update mocha dependency, update version number to 0.5.0

daschult:
I'm getting things ready for Storage to generate TypeScript SDKs, and publishing the runtime so that it's easier for storage to consume it seems like a good idea to me.


daschult:
When I ran `npm install`, npm informed me that mocha had some vulnerabilities in it. I know it's just devDependencies, but I figured I'd update it anyways. Webpack had a bunch too, but I've had trouble updating webpack in the past, so I just left that as it was.

PR #81, opened 2018-05-10
Reorganize tests

RikkiGibson:
Organizes tests in a similar way as they were in the httpPipeline branch.


RikkiGibson:
For some reason, spawning a subprocess is now causing mocha tests to hang. I believe this approach was working back in the httpPipeline branch, so I'll have to look into it around the time we start needing to serve resources from the local server.

PR #82, opened 2018-05-11
Add and URLBuilder and tests

daschult:
Supporting storage's dataplane swagger specification requires some tricky URL manipulation. When we implemented Java's runtime, the easiest way to do this was to create a URL parser/manipulator and to use that. This is just a port of the URLBuilder we used in the Java runtime and its tests.


RikkiGibson:
Is the generator going to start referencing this?

daschult:
@RikkiGibson Yes.

lib/url.ts
```diff
@@ -0,0 +1,512 @@
...
+  }
+
+  public static parse(text: string): URLBuilder {
+    const result = new URLBuilder();
+    result.set(text, URLTokenizerState.SCHEME_OR_HOST);
+    return result;
+  }
+}
+
+export enum URLTokenizerState {
```
RikkiGibson: Really small thing but may want to consider using [const enums](https://basarat.gitbooks.io/typescript/content/docs/enums.html#const-enums) for space

RikkiGibson: In fact don't even worry about that in this PR.. let's think about space stuff down the line

daschult: Too late. I've already done it. I'm just running it through my tests to make sure everything still works.

PR #83, opened 2018-05-11
Add the ability to read, modify, and substitute URL queries

daschult:
I forgot to add URL substitution functions for updating path and query arguments.

### DIFF
--- a/lib/url.ts
+++ b/lib/url.ts
@@ -1,8 +1,56 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-export interface URLQuery {
-  [queryParameterName: string]: string;
+export class URLQuery {
+  private readonly _rawQuery: { [queryParameterName: string]: string } = {};
+
+  public any(): boolean {
+    return Object.keys(this._rawQuery).length > 0;
+  }
+
+  public set(parameterName: string, parameterValue: string | undefined): void {
+    if (parameterName) {
+      if (parameterValue) {
+        this._rawQuery[parameterName] = parameterValue;
+      } else {
+        delete this._rawQuery[parameterName];
+      }
+    }
+  }
+
+  public get(parameterName: string): string | undefined {
+    return parameterName ? this._rawQuery[parameterName] : undefined;
+  }
+
+  public toString(): string {
+    let result = "";
+    for (const parameterName in this._rawQuery) {
+      if (result) {
+        result += "&";
+      }
+      result += `${parameterName}=${this._rawQuery[parameterName]}`;
+    }
+    return result;
+  }
+
+  public static parse(text: string): URLQuery {
+    const result = new URLQuery();
+
+    if (text) {
+      if (text.startsWith("?")) {
+        text = text.substring(1);
+      }
+      const queryParameters: string[] = text.split("&");
+      for (const queryParameter of queryParameters) {
+        const queryParameterParts: string[] = queryParameter.split("=");
+        if (queryParameterParts.length === 2) {
+          result.set(queryParameterParts[0], queryParameterParts[1]);
+        }
+      }
+    }
+
+    return result;
+  }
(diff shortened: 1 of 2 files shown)


lib/url.ts
```diff
@@ -1,8 +1,56 @@
...
+  private readonly _rawQuery: { [queryParameterName: string]: string } = {};
+
+  public any(): boolean {
+    return Object.keys(this._rawQuery).length > 0;
+  }
+
+  public set(parameterName: string, parameterValue: string | undefined): void {
+    if (parameterName) {
+      if (parameterValue) {
+        this._rawQuery[parameterName] = parameterValue;
```
RikkiGibson: any potential issues related to meaningful differences between an empty and an undefined parameter value?

lib/url.ts
```diff
@@ -1,8 +1,56 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-export interface URLQuery {
-  [queryParameterName: string]: string;
+export class URLQuery {
```
amarzavery: jsdoc comments for the class, properties and methods are missing. It is a good idea to document them when everything is fresh in your mind.

lib/url.ts
```diff
@@ -1,8 +1,56 @@
...
+  private readonly _rawQuery: { [queryParameterName: string]: string } = {};
+
+  public any(): boolean {
+    return Object.keys(this._rawQuery).length > 0;
+  }
+
+  public set(parameterName: string, parameterValue: string | undefined): void {
+    if (parameterName) {
+      if (parameterValue) {
```
amarzavery: what happens if the parameterValue is an empty string ? Empty string are falsy values.

lib/url.ts
daschult: I thought about it and decided to keep things simple by treating an undefined parameter value the same as an empty parameter value (both mean a parameter that doesn't exist or won't exist anymore). If we change our minds in the future, it will be pretty easy to change this.

lib/url.ts
daschult: Yup. In that case I'll remove the parameter from the URLQuery.

lib/url.ts
RikkiGibson: I think AutoRest tests will wind up proving whichever way this needs to go.

lib/url.ts
```diff
@@ -96,25 +144,62 @@ export class URLBuilder {
...
+  public getQueryParameterValue(queryParameterName: string): string | undefined {
+    return this._query ? this._query.get(queryParameterName) : undefined;
+  }
+
+  /**
+   * Get the query in this URL.
+   */
+  public getQuery(): string | undefined {
+    return this._query ? this._query.toString() : undefined;
+  }
```
amarzavery: Copying from [swagger specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-7)


collectionFormat | string | Determines the format of the array if type array is used. Possible values are:csv - comma separated values foo,bar.ssv - space separated values foo bar.tsv - tab separated values foo\tbar.pipes - pipe separated values foo\|bar.multi - corresponds to multiple parameter instances instead of multiple values for a single instance foo=bar&foo=baz. This is valid only for parameters in "query" or "formData".Default value is csv.
-- | -- | --

daschult: Since this is swagger specific, I'd like to keep it outside of the URL logic. I think that the URL classes should keep things simple and in that regard should think about query parameters as just being string to string mappings. Any formatting/conversion should happen outside of these URL specific classes.

lib/url.ts
daschult: We talked about this, and since an empty string is a valid parameter value, I'll update this logic so that it checks if the parameter value is undefined/null instead.

lib/url.ts
daschult: Added.

PR #84, opened 2018-05-14
Don't expose fetch types in HttpOperationResponse

RikkiGibson:
- Pulls HttpHeaders class as-is from httpPipeline branch
- Simplifies msRest.ts because I found out `export { ... } from "other-module"` is a thing

- Is there a reason for HttpOperationResponse itself to be a class type? Seems like maybe not since any HttpClient implementation can implement the interface with their own class type. If you don't agree I can change the interface I have there to be some "properties" type which is used as a constructor argument.

- Thinking of just providing the HttpOperationResponse directly in the generated code in the streaming case. Any apparent reason to do it differently?


RikkiGibson:
BTW corresponding PRs need to be opened in ms-rest-azure-js and autorest.typescript because this is a breaking change.

PR #86, opened 2018-05-15
Add OperationSpec with HttpMethod property

daschult:


lib/operationSpec.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { HttpMethods } from "./webResource";
+
+/**
+ * A specification that defines a single Azure service operation.
```
RikkiGibson: Azure?

daschult: Good point. Removed.

lib/operationSpec.ts
```diff
@@ -0,0 +1,14 @@
...
+import { HttpMethods } from "./webResource";
+
+/**
+ * A specification that defines an operation.
+ */
+export interface OperationSpec {
+  /**
+   * The HTTP method that should be used by requests for this operation.
+   */
+  httpMethod: HttpMethods;
```
amarzavery: Just a thought: Do you want to have an index signature `[prop: string]: any;` so that someone can provide unknown/additional properties to the operation spec and pass it to the sendRequest method.

This could be useful while you are developing/designing the operation spec.

lib/serviceClient.ts
```diff
@@ -136,6 +141,11 @@ export class ServiceClient {
     } catch (error) {
       return Promise.reject(error);
     }
+
+    if (operationSpec) {
```
amarzavery: I am little confused over here. Wouldn't the options have the http method in it. I believe RequestPrepareOptions had that. Why would someone provide the http method in the operationSpec then?

Do you want to start with a completely new method named sendRequestWithOperationSpec() rather than touching this one?

lib/operationSpec.ts
daschult: To be brief, no, I don't think an index signature would be beneficial here.
I reserve index signatures for when an interface represents a map/dictionary type where the values don't mean anything to my code, but will mean something to my customer's code. In this scenario, OperationSpec is a set of properties that will populate a provided HttpRequest. If the customer added extra properties to the OperationSpec, my code in sendRequest() wouldn't know what those properties were, and even if it did, it wouldn't know what to do with them.

lib/serviceClient.ts
amarzavery: Thus the generic sendRequest can stay as-is and will be simple to use for customers and the other flavor will be a dedicated one for operation spec. It will save us from bunch of if-else clauses (which always turn out to be potential disasters, once things get complex).

daschult: `options` (aka `httpRequest`) may have the HTTP method on it. We really have two scenarios here:
1. sendRequest() is invoked by AutoRest-generated code, and
2. sendRequest() is invoked by customer code that wasn't generated by AutoRest.
My plan is for all of our AutoRest-generated code to just pass the arguments provided to the AutoRest-generated method plus an OperationSpec to ServiceClient.sendRequest() and for sendRequest() to create the HttpRequest and to send it off. However, I can't really do that yet because I'm trying to take this in small steps.
It might make a lot of sense to create an additional method called `sendOperationRequest()` separate from `sendRequest()`, but I'm planning on holding off on that until I have pushed all of the logic in our generated methods into the ServiceClient.sendRequest() method.

lib/serviceClient.ts
```diff
@@ -161,6 +156,17 @@ export class ServiceClient {
     }
     return Promise.resolve(operationResponse);
   }
+
+  /**
+   * Send an HTTP request that is populated using the provided OperationSpec.
+   * @param {WebResource} httpRequest - The HTTP request to populate and then to send.
+   * @param {operationSpec} operationSpec - The OperationSpec to use to populate the httpRequest.
+   */
+  async sendOperationRequest(httpRequest: WebResource, operationSpec: OperationSpec): Promise<HttpOperationResponse> {
```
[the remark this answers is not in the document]
RikkiGibson: I expect the operationSpec to be immutable and shared between calls, and to be accompanied in sendOperationRequest by some arguments array or some other value that basically contains the raw arguments provided to the generated method.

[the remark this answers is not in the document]
RikkiGibson: But that's entirely guesswork, we will see how exactly things shake out :)

PR #88, opened 2018-05-15
Attempt to fix VSTS CI jobs by changing build and test npm scripts to…

daschult:
… run sequentially instead of in parallel


RikkiGibson:
Can you use GitHub references to check this hypothesis without needing to merge?

PR #90, opened 2018-05-16
Add publish scripts

RikkiGibson:
Copied from https://github.com/Azure/perks.extension


RikkiGibson:
So this is supposed to publish to the `@preview` tag when we merge.

PR #95, opened 2018-05-18
Add serializationPolicy

daschult:
Storage is very interested in having a serialization RequestPolicy, so this change adds the SerializationPolicy class and also adds it to the default HTTP pipeline.
Oh, and I have to apologize too. I accidentally committed and pushed to master earlier (I thought I was on the daschult/serialization branch), but since my changes didn't break anything, I just left it as it was. I'm going to add an issue for making our master branches protected and then we can talk about it later.


lib/policies/serializationPolicy.ts
```diff
@@ -48,8 +48,8 @@ export class SerializationPolicy extends BaseRequestPolicy {
       const bodyMapper: Mapper | undefined = operationSpec.requestBodyMapper;
       if (bodyMapper) {
         try {
-          if (request.body != undefined) {
-            request.body = this._serializer.serialize(bodyMapper, request.body, "");
+          if (request.body != undefined && operationSpec.requestBodyName) {
```
amarzavery: Make sure this covers scenarios where the request body is null. Basically someone wants to send null on the wire and it should be possible to do that.

daschult: How is null different than an empty body? Do you mean that they want to send the text string "null"?

daschult: It passes all of the tests in the autorest.typescript, which I'm guessing means it passes this request.

RikkiGibson: I'm interpreting it to mean that down the line if someone wants to encode the body object `null` as JSON, then we should translate to the string `"null"` so that 4-character sequence will be written to the wire.

You might be able to do this just by changing `request.body != undefined` to `!==`. But I'm assuming you made some efforts to preserve the way the current serializer works. If we decide to handle this use case let's start by adding a failing test 👍

RikkiGibson:
Will the generator or HttpClient have to change to stop making the call to deserialize the response bodies?

daschult:
That's coming. Right now I'm just focusing on the request bodies.

PR #99, opened 2018-05-22
Add OperationArguments and evaluate header parameters

daschult:
This PR adds header parameter evaluation to the runtime.


lib/serviceClient.ts
```diff
@@ -169,10 +175,60 @@ export class ServiceClient {
...
 
+    if (operationSpec.headerParameters) {
+      for (const headerParameter of operationSpec.headerParameters) {
+        const parameterName: string = headerParameter.parameterName;
+        const headerName: string = headerParameter.headerName || parameterName;
+        const parameterType: OperationParameterType | undefined = headerParameter.type;
+
+        let headerValue: any = operationArguments.arguments[parameterName];
+        if (headerValue != undefined) {
+          if (parameterType != undefined) {
```
RikkiGibson: does a missing parameterType indicate a Swagger/generation time error?

lib/operationArguments.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * The arguments that were passed to the operation method.
+   */
+  arguments: { [parameterName: string]: any };
+
+  /**
+   * Headers that will be applied to this operation's HTTP request after the operation method's
+   * header arguments are added.
+   */
+  customHeaders?: HttpHeaders;
```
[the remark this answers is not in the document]
daschult: No. I believe RequestOptionsBase will probably stay as it is. This is just an internal class used to pass arguments to ServiceClient.sendOperationRequest().

lib/serviceClient.ts
amarzavery: i think when we eventually go the parameter mapper route. If a required parameter is not provided then it will throw an error during serialization.

PR #100, opened 2018-05-22
Retry policies should clone HTTP requests

daschult:


lib/policies/systemErrorRetryPolicy.ts
```diff
@@ -115,10 +115,10 @@ export class SystemErrorRetryPolicy extends BaseRequestPolicy {
       // If previous operation ended with an error and the policy allows a retry, do that
       try {
         await utils.delay(retryData.retryInterval);
-        const res: HttpOperationResponse = await this._nextPolicy.sendRequest(operationResponse.request);
-        return self.retry(res, retryData, err);
+        const res: HttpOperationResponse = await this._nextPolicy.sendRequest(request.clone());
```
RikkiGibson: Is this a redundant clone? (also cloning in sendRequest)

lib/webResource.ts
```diff
@@ -231,6 +231,18 @@ export class WebResource {
...
+
+  /**
+   * Clone this WebResource HTTP request object.
+   * @returns {WebResource} The clone of this WebResource HTTP request object.
+   */
+  clone(): WebResource {
+    const result = new WebResource(this.url, this.method, this.body, this.query, this.headers, this.rawResponse, this.abortSignal);
+    result.formData = this.formData;
+    result.operationSpec = this.operationSpec;
+    result.abortSignal = this.abortSignal;
```
RikkiGibson: redundant?

lib/policies/systemErrorRetryPolicy.ts
daschult: No, we need to clone for each attempt to send the request.

RikkiGibson: Oh, you're right. I see the request passed to sendRequest on 118 is completely independent from the retry.

lib/webResource.ts
daschult: Removed in https://github.com/Azure/ms-rest-js/pull/100/commits/5fd6b70d34d2521918ec6630e02a13288f41c69f.

PR #101, opened 2018-05-22
Add HTTP request cloning to rpRegistrationPolicy

daschult:


lib/policies/rpRegistrationPolicy.ts
```diff
@@ -18,21 +18,20 @@ export class RPRegistrationPolicy extends BaseRequestPolicy {
...
 
   public async sendRequest(request: WebResource): Promise<HttpOperationResponse> {
-    const response: HttpOperationResponse = await this._nextPolicy.sendRequest(request);
-    return this.registerIfNeeded(response);
+    const response: HttpOperationResponse = await this._nextPolicy.sendRequest(request.clone());
+    return this.registerIfNeeded(request, response);
   }
 
-  async registerIfNeeded(response: HttpOperationResponse): Promise<HttpOperationResponse> {
+  async registerIfNeeded(request: WebResource, response: HttpOperationResponse): Promise<HttpOperationResponse> {
```
RikkiGibson: Did you add the parameter because you don't want to use the request that's already been sent, and instead use the "unconsumed" request?

daschult: Yes. It's necessary. It's very likely that the request that has already been sent was modified after it left this rpRegistrationPolicy (if it goes through a serializationPolicy after this policy, for example). I need the original request so that I can get back to the state the request was in when it entered the rpRegistrationPolicy.

PR #105, opened 2018-05-24
Add url construction to ServiceClient.sendOperationRequest()

daschult:


lib/operationQueryParameter.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A parameter for an operation that will be added as a query parameter to the operation's HTTP
+ * request.
+ */
+export interface OperationQueryParameter {
+  /**
+   * The name of the parameter.
+   */
+  parameterName: string;
```
RikkiGibson: what's the case where parameterName and queryParameterName varies? Does that come down to the Swagger?

lib/operationSpec.ts
```diff
@@ -14,6 +16,18 @@ export interface OperationSpec {
    */
   httpMethod: HttpMethods;
 
+  /**
+   * The URL that was provided in the service's specification. This will still have all of the URL
+   * placeholder variables in it.
```
RikkiGibson: I'd be tempted to use the word "template" over "placeholder"

lib/operationQueryParameter.ts
daschult: One common case I saw in the generator was `apiVersion` and `api-version`. There are also cases where the parameter name is `this.client.value`. I can probably optimize the later example so that it just turns into `value`, but for now I'm trying to keep it simple.

lib/operationSpec.ts
daschult: Sure. I'll change it.

lib/serviceClient.ts
```diff
@@ -189,60 +191,127 @@ export class ServiceClient {
...
+              if (queryParameterValue.length === 0) {
+                queryParameterValue = "";
+              } else {
+                for (const item of queryParameterValue) {
+                  queryParameterValue = (item == undefined ? "" : item.toString());
+                }
+              }
+            } else {
+              let queryParameterValueSeparator: string;
+              switch (queryParameter.collectionFormat) {
```
RikkiGibson: Can this be baked into the QueryCollectionFormat enum?

daschult: I can't make the enum a string enum (if that's what you're asking), but I can move this logic into the same file.

RikkiGibson: Yeah, I think that's adequate. That's because you want to fail for multi, basically?

daschult: Correct.

PR #106, opened 2018-05-24
Add abortSignal param to RequestOptionsBase

RikkiGibson:


lib/axiosHttpClient.ts
```diff
@@ -85,8 +85,9 @@ export class AxiosHttpClient implements HttpClient {
       throw new RestError("The request was aborted", "REQUEST_ABORTED_ERROR", undefined, httpRequest);
     }
 
+    let abortListener: (() => void) | undefined;
     const cancelToken = abortSignal && new axios.CancelToken(canceler => {
-      abortSignal.addEventListener("abort", () => canceler());
+      abortSignal.addEventListener("abort", abortListener = () => canceler());
```
daschult: I'm always wary of using assignments as expressions. Can you move this assignment to line 88 instead?

daschult: Or at least to a previous line.

PR #107, opened 2018-05-24
Add contentType to OperationSpec

daschult:


lib/operationSpec.ts
```diff
@@ -28,6 +28,12 @@ export interface OperationSpec {
    */
   path?: string;
 
+  /**
+   * The content type of the request body. This value will be used as the "Content-Type" header if
+   * it is provided.
+   */
+  contentType?: string;
```
RikkiGibson: if content-type is specified in customHeaders, does the custom header win?

amarzavery: That is the right approach. I remember having to do a modification in the nodejs code generator some time last year to satisfy this requirement. We would set the custom headers first in the request object. Then if the Content-Type is not set then we would set it to what was specified in the swagger spec.

PR #108, opened 2018-05-24
Enable browser tests with npm test

RikkiGibson:


package.json
```diff
@@ -25,7 +25,7 @@
     "autorest",
     "clientruntime"
   ],
-  "main": "./dist/lib/msRest.js",
+  "main": "./lib/msRest.ts",
```
RikkiGibson: revert

test/shared/axiosHttpClientTests.ts
```diff
@@ -96,11 +96,8 @@ describe("axiosHttpClient", () => {
   });
 
   it("should allow canceling requests", async function() {
-    // ensure that a large upload is actually cancelled
-    this.timeout(2000);
-
     const controller = getAbortController();
-    const request = new WebResource(`${baseURL}/fileupload`, "POST", new Uint8Array(1024*1024*100), undefined, undefined, true, controller.signal);
+    const request = new WebResource(`${baseURL}/fileupload`, "POST", new Uint8Array(1024*1024*10), undefined, undefined, true, controller.signal);
```
RikkiGibson: These sizes make mocha-chrome freak out and take 14 seconds to abort :/ Will want to do some stress testing separately.

.scripts/chrome-unit.js
```diff
@@ -0,0 +1,21 @@
...
+    const mochaChrome = spawn(join(__dirname, "../node_modules/.bin/mocha-chrome"), ["http://localhost:3001"], { shell: true, stdio: "inherit" });
+    mochaChrome.on("exit", () => {
+      webpackDevServer.stderr.destroy();
+      webpackDevServer.stdout.destroy();
+      webpackDevServer.kill();
+    });
+  }
+}
+
+webpackDevServer.stderr.on('data', data => console.error(data.toString()));
```
daschult: Could you put this into the webpackDevServerHandler function/lambda so that you only have a single 'data' handler?

test/shared/axiosHttpClientTests.ts
amarzavery: Do file an issue for that for tracking purpose.

RikkiGibson: #109

.scripts/chrome-unit.js
RikkiGibson: I think I do I want to log errors that come from the dev server, but not log most of the regular output (i.e. the webpack build report).

RikkiGibson:
@amarzavery do you foresee any possible problems with using `sudo: true` in order to make headless Chrome work properly in Travis?

RikkiGibson:
@daschult @amarzavery We are finally green on automated Chrome tests! I may want to add the same to autorest.typescript. I wouldn't expect it to add more than 30 seconds to the build..

daschult:
@RikkiGibson Congratulations! 30 seconds should be fine, especially if you can get the tests to run in parallel.

RikkiGibson:
Just got:

```sh
Error: listen EADDRINUSE :::3001
```

Might have to go back to a straight-up `unit` script. That will let us launch the testserver once and parallelize, at least.

RikkiGibson:
*sigh* this disrupted the publish task in VSTS, because yarn wants to run tests before publishing. Will fiddle with it and try to fix now.

PR #112, opened 2018-05-25
Don't overwrite cookie if already present on request

### LINKED ISSUE #89: Unable to set cookie header
I'm dealing with a funky authentication process that requires me to:
1. Hit an endpoint with redirects set to 0 or handle all responses in the chain
2. Take the `Set-Cookie` header value from the first response and supply it with all future requests in the `Cookie` header

I'm attempting to accomplish this through a custom implementation of `ServiceClientCredentials` utilizing Axios(for now at least). For the life of me, whatever I try, the `cookie` header ends up blank on the wire. It appears some effort was made to get cookies working with fetch, however I can't get them working for myself :)

Any guidance on this would be much appreciated(or if I could convince you to utilize Axios :D ).

RikkiGibson:
Resolves #89 (hopefully, if I'm understanding the problem correctly)

### DIFF
--- a/test/shared/axiosHttpClientTests.ts
+++ b/test/shared/axiosHttpClientTests.ts
@@ -117,6 +117,16 @@ describe("axiosHttpClient", () => {
     }
   });
 
+  it("should not overwrite a user-provided cookie", async function() {
+    const client = new AxiosHttpClient();
+    const request1 = new WebResource(`${baseURL}/set-cookie`);
+    await client.sendRequest(request1);
+
+    const request2 = new WebResource(`${baseURL}/cookie`, "GET", undefined, undefined, { Cookie: "data=abcdefg" });
+    const response = await client.sendRequest(request2);
+    should(response.headers.get("Cookie")).equal("data=abcdefg");
+  });
+
   it("should allow canceling multiple requests with one token", async function () {
     // ensure that a large upload is actually cancelled
     this.timeout(4000);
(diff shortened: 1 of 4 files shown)


RikkiGibson:
Please feel free to chime in @ProTip as to whether this seems like the solution you need or not (and feel free to try just changing your ms-rest-js version to `RikkiGibson/ms-rest-js#RespectUserCookie` to see if it fixes your problem).

test/shared/axiosHttpClientTests.ts
```diff
@@ -117,6 +117,16 @@ describe("axiosHttpClient", () => {
     }
   });
 
+  it("should not overwrite a user-provided cookie", async function() {
+    const client = new AxiosHttpClient();
+    const request1 = new WebResource(`${baseURL}/set-cookie`);
+    await client.sendRequest(request1);
+
+    const request2 = new WebResource(`${baseURL}/cookie`, "GET", undefined, undefined, { Cookie: "data=abcdefg" });
```
daschult: Maybe add another `/cookie` request here without a Cookie header that demonstrates that the "Cookie": "data=123456" cookie is in the response headers. That would contrast the current second request where the "Cookie": "data=abcdefg" header overrides the existing cookie in the cookieJar.

ProTip:
@RikkiGibson Thanks! I'm actually in the process of testing this right now. The build process seems to be rather opaque ATM..  Is the code published to npm under a certain tag? Do I need to clone, build, and link the project to test?

PR #115, opened 2018-05-30
Add onUpload/DownloadProgress properties and createOperationArguments

RikkiGibson:
Leaving aside the rename/constructor refactor work for now as it's not trivial to unify `constructor()` and `prepare()` and will break people.

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -103,6 +103,12 @@ export class AxiosHttpClient implements HttpClient {
       bodyType === "function" ? httpRequest.body() :
       httpRequest.body;
 
+    const userUploadProgress = httpRequest.onUploadProgress;
+    const onUploadProgress = userUploadProgress && ((rawEvent: ProgressEvent) => userUploadProgress({ loaded: rawEvent.loaded, total: rawEvent.lengthComputable ? rawEvent.total : undefined }));
+
+    const userDownloadProgress = httpRequest.onDownloadProgress;
+    const onDownloadProgress = userDownloadProgress && ((rawEvent: ProgressEvent) => userDownloadProgress({ loaded: rawEvent.loaded, total: rawEvent.lengthComputable ? rawEvent.total : undefined }));
+
     let res: AxiosResponse;
     try {
       const config: AxiosRequestConfig = {
@@ -116,7 +122,9 @@ export class AxiosHttpClient implements HttpClient {
         // Workaround for https://github.com/axios/axios/issues/1362
         maxContentLength: 1024 * 1024 * 1024 * 10,
         responseType: httpRequest.rawResponse ? (isNode ? "stream" : "blob") : "text",
-        cancelToken
+        cancelToken,
+        onUploadProgress,
+        onDownloadProgress
       };
       res = await axiosClient(config);
     } catch (err) {
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -19,6 +19,8 @@ if (isNode) {
   axiosClient.interceptors.request.use(config => ({ ...config, method: config.method && config.method.toUpperCase() }));
 }
 
+type AxiosProgressFunction = (rawEvent: ProgressEvent) => void;
+
 /**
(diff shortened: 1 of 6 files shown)


lib/operationArguments.ts
```diff
@@ -20,4 +22,30 @@ export interface OperationArguments {
...
+/**
+ * Assigns the properties of a RequestOptions object to an OperationArguments object.
+ * @param args the arguments object to use in the OperatonArguments
+ * @param options the RequestOptions to apply
+ * @return an OperationArguments object
+ */
+export function createOperationArguments(args: { [parameterName: string]: any }, options?: RequestOptionsBase): OperationArguments {
+  return {
+    arguments: args,
+    customHeaders: options && options.customHeaders,
```
RikkiGibson: may want to just check `if(options) {...}` and assign this stuff.. It's a small code size/possible(?) performance change.

RikkiGibson: I'm optimistic that this function will reduce the code size a bit instead of increasing it by adding onUpload/Download progress property initializers inline in the generated code.

lib/axiosHttpClient.ts
```diff
@@ -103,6 +103,12 @@ export class AxiosHttpClient implements HttpClient {
       bodyType === "function" ? httpRequest.body() :
       httpRequest.body;
 
+    const userUploadProgress = httpRequest.onUploadProgress;
+    const onUploadProgress = userUploadProgress && ((rawEvent: ProgressEvent) => userUploadProgress({ loaded: rawEvent.loaded, total: rawEvent.lengthComputable ? rawEvent.total : undefined }));
+
+    const userDownloadProgress = httpRequest.onDownloadProgress;
+    const onDownloadProgress = userDownloadProgress && ((rawEvent: ProgressEvent) => userDownloadProgress({ loaded: rawEvent.loaded, total: rawEvent.lengthComputable ? rawEvent.total : undefined }));
```
daschult: Can you add type definitions for these variables? Adding the raw type could be bad, so I'd recommend adding typedefs at the top of this file (UploadProgressFunction and DownloadProgressFunction) and then using them here.

lib/webResource.ts
```diff
@@ -9,6 +9,22 @@ import { generateUuid } from "./util/utils";
...
+  /**
+   * The number of bytes loaded so far.
+   */
+  loaded: number,
+
+  /**
+   * The total number of bytes that will be loaded.
+   * If the total number of bytes is unknown, this property will be undefined.
+   */
+  total?: number
```
daschult: I vote that this be called `totalBytes`. I'm almost always in favor of including a variable's units in its name.

lib/webResource.ts
```diff
@@ -9,6 +9,22 @@ import { generateUuid } from "./util/utils";
...
 export type HttpRequestBody = Blob | string | ArrayBuffer | ArrayBufferView | (() => NodeJS.ReadableStream);
 
+/**
+ * Fired in response to upload or download progress (browser only).
+ */
+export type TransferProgressEvent = {
+  /**
+   * The number of bytes loaded so far.
+   */
+  loaded: number,
```
daschult: Were there any other names you considered besides `loaded`?

RikkiGibson: I just used the names from the underlying [ProgressEvent](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent).

lib/axiosHttpClient.ts
RikkiGibson: Do you mean like `const onUploadProgress: ProgressFunction = ...`?

lib/webResource.ts
RikkiGibson: So `loadedBytes` and `totalBytes` might be better?

lib/axiosHttpClient.ts
daschult: Yes.

lib/webResource.ts
daschult: I prefer those names. They remind me that we're dealing with bytes, without having to lookup the documentation.

PR #116, opened 2018-05-30
Show output when test process fails

RikkiGibson:
Got a bit ahead of myself-- we need to always log the output from the test processes so we can see the failures.


.scripts/unit.js
```diff
@@ -49,7 +49,6 @@ const webpackDevServerHandler = (data) => {
       cleanupDevServer();
       process.exit(0);
     }).catch((err) => {
-      console.error(err);
```
RikkiGibson: This error message winds up being useless--basically telling us what we already know from the subprocess output

PR #117, opened 2018-05-31
Use ImportHelpers setting

RikkiGibson:
This brings our bundle from 93.2KB -> 81.4KB. Just seemed so easy to add in that I should just do it. Will do the same with ms-rest-azure-js.


RikkiGibson:
Addresses Azure/autorest.typescript#126

PR #121, opened 2018-05-31
Parse response body when charset is appended to Content-Type

### LINKED ISSUE #114: Response data not parsed when content type includes encoding
I'm getting back a content type that includes encoding: ```application/json;charset=UTF-8```.

This is failing the check in the axios client on ```@preview```:
```js
else if (contentType === "application/json" || contentType === "text/json" || !contentType)
```

I do not get parsed data back :|

RikkiGibson:
Fixes #114

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -181,8 +181,8 @@ export class AxiosHttpClient implements HttpClient {
 
       try {
         if (operationResponse.bodyAsText) {
-          const contentType = operationResponse.headers.get("Content-Type");
-          if (contentType === "application/xml" || contentType === "text/xml") {
+          const contentType = operationResponse.headers.get("Content-Type") || '';
+          if (contentType.includes("application/xml") || contentType.includes("text/xml")) {
             const xmlParser = new xml2js.Parser(XML2JS_PARSER_OPTS);
             const parseString = new Promise(function (resolve: (result: any) => void, reject: (err: any) => void) {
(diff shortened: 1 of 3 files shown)


RikkiGibson:
Thanks for your patience @ProTip @KyleGobel

lib/axiosHttpClient.ts
```diff
@@ -181,8 +181,8 @@ export class AxiosHttpClient implements HttpClient {
 
       try {
         if (operationResponse.bodyAsText) {
-          const contentType = operationResponse.headers.get("Content-Type");
-          if (contentType === "application/xml" || contentType === "text/xml") {
+          const contentType = operationResponse.headers.get("Content-Type") || '';
+          if (contentType.includes("application/xml") || contentType.includes("text/xml")) {
```
daschult: `includes()` is a dangerous thing to use here because it will match `application/xmlblah`, which may not be desired. I really recommend splitting this and doing an equality check instead.

RikkiGibson: Please have a look at 4b1ddc5

lib/axiosHttpClient.ts
```diff
@@ -181,8 +181,11 @@ export class AxiosHttpClient implements HttpClient {
 
       try {
         if (operationResponse.bodyAsText) {
-          const contentType = operationResponse.headers.get("Content-Type");
-          if (contentType === "application/xml" || contentType === "text/xml") {
+          const contentType = operationResponse.headers.get("Content-Type") || "";
+          console.log(contentType);
```
RikkiGibson: Remove

PR #122, opened 2018-05-31
Add support for parameter mappers

daschult:
ms-rest-js changes necessary for https://github.com/Azure/autorest.typescript/issues/94.


lib/serializer.ts
```diff
@@ -156,10 +158,10 @@ export class Serializer {
       } else if (typeName.match(/^Stream$/ig) !== null) {
         const objectType = typeof value;
         if (objectType !== "string" &&
-            objectType !== "function" &&
-            !(value instanceof ArrayBuffer) &&
-            !ArrayBuffer.isView(value) &&
-            !(typeof Blob === "function" && value instanceof Blob)) {
+          objectType !== "function" &&
```
RikkiGibson: I'd prefer if this either were reindented or a blank line were inserted above the `throw` statement in order to make it clear where the condition ends and the body begins.

lib/serializer.ts
```diff
@@ -729,7 +731,7 @@ export interface Mapper {
   xmlIsWrapped?: boolean;
   readOnly?: boolean;
   isConstant?: boolean;
-  required: boolean;
+  required?: boolean;
```
RikkiGibson: poetic

test/shared/axiosHttpClientTests.ts
```diff
@@ -3,9 +3,9 @@
 import * as assert from "assert";
 import * as should from "should";
 import { AxiosHttpClient } from "../../lib/axiosHttpClient";
-import { baseURL } from "../testUtils";
+import { isNode } from "../../lib/util/utils";
```
RikkiGibson: sorting?

lib/serializer.ts
daschult: I'll insert the blank line. I agree with you about the indentation, but the VS Code TypeScript formatter made it this way.

test/shared/axiosHttpClientTests.ts
daschult: According to the VS Code TypeScript extension's import sorting algorithm, this is the correct sorting order.

PR #124, opened 2018-06-02
Remove usages of Buffer

RikkiGibson:
This is bringing back work done in the past, maybe done on the httpPipeline branch and forgotten. This should save about 20KB when bundling for the browser. I believe there might be corresponding work to do in the generator.

### DIFF
--- a/lib/util/base64.ts
+++ b/lib/util/base64.ts
@@ -19,7 +19,11 @@ export function encodeString(value: string): string {
  */
 export function encodeByteArray(value: Uint8Array): string {
   if (typeof Buffer === "undefined") {
-    return btoa(encodeURIComponent(new TextDecoder().decode(value)));
+    let str = "";
+    for (let i = 0; i < value.length; i++) {
+      str += String.fromCharCode(value[i]);
+    }
+    return btoa(str);
   } else {
     // Buffer.from accepts <ArrayBuffer> | <SharedArrayBuffer>-- the TypeScript definition is off here
     // https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length
@@ -34,7 +38,12 @@ export function encodeByteArray(value: Uint8Array): string {
  */
 export function decodeString(value: string): Uint8Array {
   if (typeof Buffer === "undefined") {
-    return new TextEncoder().encode(atob(value));
+    const byteString = atob(value);
+    const arr = new Uint8Array(byteString.length);
+    for (let i = 0; i < byteString.length; i++) {
+      arr[i] = byteString.charCodeAt(i);
+    }
+    return arr;
   } else {
     return Buffer.from(value, "base64");
   }
--- a/lib/util/base64.ts
+++ b/lib/util/base64.ts
@@ -19,9 +19,11 @@ export function encodeString(value: string): string {
  */
 export function encodeByteArray(value: Uint8Array): string {
   if (typeof Buffer === "undefined") {
-    return btoa(new TextDecoder().decode(value));
+    return btoa(encodeURIComponent(new TextDecoder().decode(value)));
   } else {
-    const bufferValue = (value instanceof Buffer) ? value : new Buffer(value);
+    // Buffer.from accepts <ArrayBuffer> | <SharedArrayBuffer>-- the TypeScript definition is off here
+    // https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length
+    const bufferValue = (value instanceof Buffer) ? value : Buffer.from(value.buffer as ArrayBuffer);
     return bufferValue.toString("base64");
   }
 }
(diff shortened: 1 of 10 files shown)


lib/util/base64.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Encodes a byte array in base64 format.
+ * @param value the Uint8Aray to encode
+ */
+export function encodeByteArray(value: Uint8Array): string {
+  if (typeof Buffer === "undefined") {
+    return btoa(new TextDecoder().decode(value));
+  } else {
+    const bufferValue = (value instanceof Buffer) ? value : new Buffer(value);
```
amarzavery: new versions of node.js [have deprecated](https://nodejs.org/dist/latest-v8.x/docs/api/buffer.html#buffer_new_buffer_array) the `new Buffer()` syntax. They expect us to use `Buffer.from()` instead.

test/shared/serializationTests.ts
```diff
@@ -10,6 +10,14 @@ import { Mappers } from "./data/TestClient/lib/models/mappers";
 let Serializer = new msRest.Serializer({});
 let valid_uuid = "ceaafd1e-f936-429f-bbfc-82ee75dddc33";
 
+function stringToByteArray(str: string): Uint8Array {
+  if (typeof Buffer === "function") {
+    return new Buffer(str, "utf-8");
```
amarzavery: `Buffer.from(str, "utf-8");`

webpack.config.ts
```diff
@@ -10,18 +10,23 @@ const config: webpack.Configuration = {
...
-        options: { configFile: 'tsconfig.webpack.json' }
+        exclude: /(node_modules|test)/
       }
     ]
   },
   resolve: {
-    extensions: [".tsx", ".ts", ".js"]
+    extensions: [".tsx", ".ts", ".js"],
+    alias: {
+      "moment": path.resolve('./node_modules/moment/min/moment.min.js')
```
amarzavery: Do we still need moment? Haven't we moved away from it?

webpack.config.ts
```diff
@@ -10,18 +10,23 @@ const config: webpack.Configuration = {
     libraryTarget: 'var',
     library: 'msRest'
   },
+  plugins: [
+    new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/)
```
amarzavery: same as below

webpack.testconfig.ts
```diff
@@ -13,21 +13,23 @@ const config = {
     filename: 'testBundle.js',
     path: __dirname
   },
+  plugins: [
+    new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/)
```
amarzavery: same as above.

amarzavery:
- Some changes need to happen in the generator as well. You can find all the places where we are using Buffer over [here](https://github.com/Azure/autorest.typescript/search?l=C%23&q=buffer).
- It would be nice if we have a function named `isBufferLike(input: any): boolean` that accepts any input and returns true if the given input is an instance of Buffer or is a UInt8Array. I like [this approach](https://github.com/feross/is-buffer/blob/master/index.js) of finding whether it is a node.js buffer. We should copy that in ms-rest-js but not expose it. `isBufferLike()` will internally call this function. If it is false then it will validate whether the given type is an `UInt8Array` and return true or false accordingly.
- We should use the above method in serializer.ts to validate whether the given object is buffer like during serialization.

amarzavery:
Once this PR is merged, we should have a PR out in the generator as well, that updates references to Buffer, updates the minimum compatible version of ms-rest-js in generated package.json and emits a warning/info log message about the compatible version. This will ensure customers don't see errors at runtime.

RikkiGibson:
I'm not ready to merge this yet. Something funny is going on with base64 encoding in the browser. It's a bit annoying since it seemed like it was working before. Maybe CI didn't catch it because browser tests weren't enabled the first time?

Also, TextEncoder/Decoder are not a thing in IE/Edge. The best thing (slightly irritatingly) may be to simply add in raw base64 encoding/decoding functions in ms-rest-js and use them in both node.js and browser.

RikkiGibson:
Alright, now after banging my head against the spec for btoa/atob for a while, and pulling in and summarily rejecting multiple third party base64 implementations, I've settled on something that I believe works.

https://html.spec.whatwg.org/multipage/webappapis.html#atob

Note:
- a is short for ascii, which refers to a base64 encoded string consisting of ascii characters.
- b is short for binary, which refers to a string consisting of values ranging from `"\u0000"` to `"\u00ff"`.
- TextDecoder wants to treat your Uint8Array as an array of UTF-8 data by default, so to make it produce a string that `btoa` will understand, you have to `new TextDecoder("ascii")`. But this API isn't present in Edge or IE, so it's kind of a moot point-- let's just create a string of bytes from the Uint8Array and carry on.
- Do the opposite when converting from binary string to Uint8Array. See `base64.decodeString`.

PR #125, opened 2018-06-05
Remove more stuff

RikkiGibson:
- Removes `url-parse` module in favor of our own `URLBuilder` to save a few kBs
  - url-parse is about 3.5 kB while URLBuilder is about 7 kB. Is it accurate @daschult that URLBuilder does stuff that url-parse can't?
- Require specifically `uuid/v4` to save a kB
- Disable `process` polyfill to save a kB
  - Don't know how to help our users do this in their own webpack config, but it's a small thing and they're likely to have other code that depends on the process polyfill, so not sure if it matters
- Replace `msRestUserAgentPolicy.js` with a stub in browser to save a few kBs
  - See the [package-browser-field-spec](https://github.com/defunctzombie/package-browser-field-spec) to understand what I was doing there
- Removes `withCredentials: true` in axios to enable more typical CORS use cases
  - Before, we were trying to send cookies over CORS in order to successfully run autorest.typescript tests in browser. We don't need that any more and `withCredentials: true` was preventing our runtime from communicating with ARM over CORS
- Output es6 in `./es/` folder to get better results from webpack tree shaking. This saves about 10 kB.

Overall, this brings msRestBundle.js from 88.3 kB to 70.1 kB.

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -122,6 +122,7 @@ export class AxiosHttpClient implements HttpClient {
         data: axiosBody,
         transformResponse: undefined,
         validateStatus: () => true,
+        maxRedirects: 0,
         withCredentials: true,
         // Workaround for https://github.com/axios/axios/issues/1362
         maxContentLength: 1024 * 1024 * 1024 * 10,
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -122,8 +122,6 @@ export class AxiosHttpClient implements HttpClient {
         data: axiosBody,
         transformResponse: undefined,
         validateStatus: () => true,
-        maxRedirects: 0,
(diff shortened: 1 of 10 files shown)


RikkiGibson:
Please have a look when you get a chance @daschult

R0nd:
With the removal of withCredentials, how to use CORS in browser? I used to pass {credentials:"include"} via requestOptions when I constructed the autorest client, now obviously that doesn't work. Couldn't find anything in docs.

RikkiGibson:
Was that back when we used fetch under the hood? We can definitely add a parameter for it--this PR just made it so we didn't hard code it with every single request, even with CORS configurations that disallow credentials e.g. `Access-Control-Allow-Origin: *`.

RikkiGibson:
Yes, please! That seems like a very reasonable request.

PR #131, opened 2018-06-11
Move parameter unflattening and parameter ungrouping to runtime

daschult:


package.json
```diff
@@ -5,7 +5,7 @@
     "email": "azsdkteam@microsoft.com",
     "url": "https://github.com/Azure/ms-rest-js"
   },
-  "version": "0.9.0",
+  "version": "0.9.1",
```
RikkiGibson: I think our publishing script will drop this patch version and use the commit-counting thing instead.

daschult: Oh, right. Good point. I'll remove this.

test/shared/serviceClientTests.ts
```diff
@@ -0,0 +1,263 @@
...
+
+import * as assert from "assert";
+import { getPropertyParent, PropertyParent, applyParameterTransformations } from "../../lib/serviceClient";
+import { OperationArguments } from "../../lib/operationArguments";
+import { Serializer } from "../../lib/serializer";
+
+describe("applyParameterTransformations()", () => {
+  it("should do nothing if provided an undefined operationArguments", () => {
+    applyParameterTransformations(
+      undefined as any,
```
RikkiGibson: are there cases when we will suspect the generator of passing values like this?

daschult: No, this should never happen. The only reason these tests are here is because I like to know how my code will behave even in these unexpected scenarios. I'm fine if the code throws an error, but I want to know what that error will be while I'm running my tests and not when my code is running on a customer's machine.

PR #140, opened 2018-06-19
Improve type safety of mappers

RikkiGibson:
This leverages the tagged unions feature to improve the type safety and maintainability of the Mapper interfaces. It solves an issue where the new headerCollectionPrefix failed to type check, and removes the `[key: string]: any` index signature that made us work around variance in the `type` property of various mappers by essentially not type checking them.

### DIFF
--- a/lib/serializer.ts
+++ b/lib/serializer.ts
@@ -8,43 +8,55 @@ export class Serializer {
   constructor(public readonly modelMappers?: { [key: string]: any }, public readonly isXML?: boolean) { }
 
   validateConstraints(mapper: Mapper, value: any, objectName: string): void {
-    const constraints = mapper.constraints;
     const failValidation = (constraintName: keyof MapperConstraints, constraintValue: any) => {
       throw new Error(`"${objectName}" with value "${value}" should satisfy the constraint "${constraintName}": ${constraintValue}.`);
     };
-    if (constraints && (value != undefined)) {
-      if (constraints.ExclusiveMaximum != undefined && value >= constraints.ExclusiveMaximum) {
-        failValidation("ExclusiveMaximum", constraints.ExclusiveMaximum);
+    if (mapper.constraints && (value != undefined)) {
+      const {
+        ExclusiveMaximum,
+        ExclusiveMinimum,
+        InclusiveMaximum,
+        InclusiveMinimum,
+        MaxItems,
+        MaxLength,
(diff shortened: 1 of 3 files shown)


RikkiGibson:
Still checking on the impact in autorest.typescript, but I think it's fine over there.

RikkiGibson:
This is compatible with autorest.typescript's type checking phase.

RikkiGibson:
Also took an opportunity to improve the constraint validation in `serializer.ts` which saves a kB or so in addition to being easier to read (IMO) and just plain more efficient in the big-O sense.

lib/serializer.ts
```diff
@@ -393,10 +379,8 @@ function serializeDictionaryType(serializer: Serializer, mapper: DictionaryMappe
       `mapper and it must of type "object" in ${objectName}.`);
   }
   const tempDictionary: { [key: string]: any } = {};
-  for (const key in object) {
-    if (object.hasOwnProperty(key)) {
-      tempDictionary[key] = serializer.serialize(mapper.type.value, object[key], objectName);
-    }
+  for (const key of Object.keys(object)) {
+    tempDictionary[key] = serializer.serialize(mapper.type.value, object[key], objectName + "." + key);
```
RikkiGibson: This lets us give better errors by telling which dictionary key failed serialization, e.g.

`myDictionary.firstName with value "Rikki&" should satisfy the constraint /[A-Za-z_]/`

instead of `myDictionary with value "Rikki&" should satisfy...`

lib/serializer.ts
```diff
@@ -752,38 +747,40 @@ export interface Mapper {
...
   [key: string]: string;
 }
 
-export interface CompositeMapper extends Mapper {
+export interface CompositeMapper extends BaseMapper {
   type: {
     name: "Composite";
     className: string;
-    modelProperties?: { [propertyName: string]: Mapper };
+    modelProperties: { [propertyName: string]: Mapper };
```
daschult: This is a breaking change. Make sure you bump the minor version.

RikkiGibson: Note that the removal of the index signature on BaseMapper type is what allows any of this to be checked in nested mappers at all--so the potential breaks go beyond this. I think the mappers we generate type-check against these interfaces before and after the recent generator changes, but I'll bump the minor version anyway.

Issue #97, opened 2018-05-21
Put response deserialization into SerializationPolicy using legacy serializer


daschult:

RikkiGibson:
This might be needed, in addition to response parsing, in order to make it reasonable for our users to implement their own HttpClient.

RikkiGibson:
If we add options or a separate factory method to include the xml parser/builder, we could make xml an a-la-carte feature in the bundle as well.

PR #141, opened 2018-06-21
Move error deserialization into serializationPolicy

daschult:
Resolves https://github.com/Azure/ms-rest-js/issues/97 (partially).

### DIFF
--- a/lib/operationResponse.ts
+++ b/lib/operationResponse.ts
@@ -16,9 +16,4 @@ export interface OperationResponse {
    * The mapper that will be used to deserialize the response body.
    */
   bodyMapper?: Mapper;
-
-  /**
-   * Whether or not the response body will be a cloud error.
-   */
-  isCloudError?: boolean;
 }
--- a/lib/operationResponse.ts
+++ b/lib/operationResponse.ts
@@ -1,0 +1,24 @@
+import { Mapper } from "./serializer";
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+/**
+ * An OperationResponse that can be returned from an operation request for a single status code.
+ */
+export interface OperationResponse {
+  /**
+   * The mapper that will be used to deserialize the response headers.
+   */
+  headersMapper?: Mapper;
+
+  /**
+   * The mapper that will be used to deserialize the response body.
+   */
+  bodyMapper?: Mapper;
+
+  /**
+   * Whether or not the response body will be a cloud error.
+   */
+  isCloudError?: boolean;
+}
--- a/lib/policies/serializationPolicy.ts
+++ b/lib/policies/serializationPolicy.ts
@@ -3,7 +3,9 @@
 
 import { HttpOperationResponse } from "../httpOperationResponse";
 import { getPathStringFromParameter } from "../operationParameter";
+import { OperationResponse } from "../operationResponse";
 import { OperationSpec } from "../operationSpec";
+import { RestError } from "../restError";
 import { Mapper, MapperType } from "../serializer";
 import * as utils from "../util/utils";
 import { WebResource } from "../webResource";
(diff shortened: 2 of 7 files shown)


lib/operationResponse.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  /**
+   * The mapper that will be used to deserialize the response body.
+   */
+  bodyMapper?: Mapper;
+
+  /**
+   * Whether or not the response body will be a cloud error.
+   */
+  isCloudError?: boolean;
```
RikkiGibson: `isError`?

daschult: I think I was able to figure out a different way to do this. I'll remove it.

lib/policies/serializationPolicy.ts
```diff
@@ -69,4 +72,88 @@ export class SerializationPolicy extends BaseRequestPolicy {
...
+                }
+                error.body = operationSpec.serializer.deserialize(defaultResponseBodyMapper, valueToDeserialize, "error.body");
+              }
+            }
+          } catch (defaultError) {
+            error.message = `Error \"${defaultError.message}\" occurred in deserializing the responseBody - \"${response.bodyAsText}\" for the default response.`;
+          }
+          return Promise.reject(error);
+        }
+      } else {
```
RikkiGibson: dead branch?

daschult: Yes, it is right now. This is where deserializing expected responses will go. I'll remove it for now.

amarzavery:
Couple of questions/comments related to the design:
1. If error deserialization happens as a serialization policy, what will happen to customers like portal who dont want to use the request pipeline? They would still need the deserialized errors. Wouldn't it be better to deserialize them in sendOperationRequest as that method has the operationspec and knows how to deserialize positive/negative responses.
2. Since we are moving this to the pipeline, would the order in which deserialization policy is applied be important. By this I mean this should be the last policy in the pipeline before the request is sent on the wire, so that this would be the first one when response is received. It will deserialize the error if an error happened and this error would then be given to retry/redirect policies. (Do you think, the design puts restrictions on how things need to be added in the pipeline?)

daschult:
@amarzavery Good questions.
1. To be clear, Portal doesn't want to use the default pipeline policies. In order to do this, they will have to provide an empty RequestPolicyCreator[] to their ServiceClientOptions object. If they decide that they want to add our serialization policy, they only have to add that single policy (serialization) and then they get what they want.
2. Yes, putting deserialization into the pipeline definitely places restrictions on how policies are ordered. If a policy depends on the deserialized response body to be present, then that policy must come before the serialization policy so that when the response is coming back through the pipeline it will encounter that policy after the serialization policy. This is actually a request that comes directly from storage. The idea is that customizing the request policy pipeline is an advanced feature, and, just like other advanced features, it can be tricky to get correct if you don't know what you're doing. That's why we provide a default policy pipeline that puts the serialization policy at the very end.
Does this answer you questions?

RikkiGibson:
Seems fine, just make sure to address the CI error

amarzavery:
> 1. To be clear, Portal doesn't want to use the default pipeline policies. In order to do this, they will have to provide an empty RequestPolicyCreator[] to their ServiceClientOptions object. If they decide that they want to add our serialization policy, they only have to add that single policy (serialization) and then they get what they want.

It should be super clear to them that this is what they are supposed to do. My fear, is that it is not obvious and people will miss doing that. May be documenting this upfront in our runtime repo (starting with a FAQ that grows over time) would be helpful.

> 2. Yes, putting deserialization into the pipeline definitely places restrictions on how policies are ordered. If a policy depends on the deserialized response body to be present, then that policy must come before the serialization policy so that when the response is coming back through the pipeline it will encounter that policy after the serialization policy. This is actually a request that comes directly from storage. The idea is that customizing the request policy pipeline is an advanced feature, and, just like other advanced features, it can be tricky to get correct if you don't know what you're doing. That's why we provide a default policy pipeline that puts the serialization policy at the very end.

Agree with this. Creating your own pipeline is an advanced feature and there would be trade-offs for doing so. I would be better if we document our default request pipeline and every policy in a markdown document and put it in the documentation folder in this repo. We could have a simple diagram of numbered boxes representing different policies and the bidirectional link between them (basically the diagram of a doubly linkedlist). In a different section we describe what those policies do in a brief text.

```
  +--------+   +-------+    +-------+
  |        | --+ ditaa +--> |       |
  |  Text  |   +-------+    |diagram|
  |Document|   |!magic!|    |       |
  |     {d}|   |       |    |       |
  +---+----+   +-------+    +-------+
      :                         ^
      |       Lots of work      |
      +-------------------------+
  ```

daschult:
@amarzavery That's a good idea. I've created https://github.com/Azure/ms-rest-js/issues/142 for adding that documentation.

RikkiGibson:
Note that the pipeline has always been singly linked--there is no way by default to reference the "previous" policy from a given "current" policy

amarzavery:
otherwise everything looks good to me.

PR #143, opened 2018-06-21
Bump the minor version

daschult:
I forgot to bump the minor version in my previous PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ms-rest-js",
-  "version": "0.10.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ms-rest-js",
-  "version": "0.10.0",
```
RikkiGibson: welp

RikkiGibson: npm doesn't publish this file by design, so it actually doesn't matter

PR #144, opened 2018-06-21
Fix the main property in package.json

daschult:
I'm just really sloppy today. Sorry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "autorest",
     "clientruntime"
   ],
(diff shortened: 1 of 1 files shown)


RikkiGibson:
It happens. Thinking of changing this behavior anyway as the bundler is going to look at either the `browser` or `module` properties, so as far as I know this on-the-fly package.json tweak doesn't get us source maps back to the original .ts files in the browser.

PR #148, opened 2018-06-22
Use XhrHttpClient in browser

### LINKED ISSUE #138: Move JSON.parse() and XML.parse() from AxiosHttpClient to serialization policy

RikkiGibson:
Fixes #138 by moving JSON/XML parsing to the serializationPolicy. This should also make possible to add variants of serializationPolicy which handle either only JSON or both JSON and XML, which should save our JSON-only users another 6 kB.

This brings our bundle size from 67 kB -> 53.5 kB.

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -43,22 +43,20 @@ export class AxiosHttpClient implements HttpClient {
           requestForm.append(key, value);
         }
       };
-      for (const formKey in formData) {
-        if (formData.hasOwnProperty(formKey)) {
-          const formValue = formData[formKey];
-          if (formValue instanceof Array) {
-            for (let j = 0; j < formValue.length; j++) {
-              appendFormValue(formKey, formValue[j]);
-            }
-          } else {
-            appendFormValue(formKey, formValue);
+      for (const formKey of Object.keys(formData)) {
+        const formValue = formData[formKey];
+        if (Array.isArray(formValue)) {
+          for (let j = 0; j < formValue.length; j++) {
+            appendFormValue(formKey, formValue[j]);
           }
+        } else {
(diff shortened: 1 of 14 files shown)


lib/axiosHttpClient.ts
```diff
@@ -105,14 +100,6 @@ export class AxiosHttpClient implements HttpClient {
       bodyType === "function" ? httpRequest.body() :
       httpRequest.body;
 
-    const userUploadProgress = httpRequest.onUploadProgress;
```
RikkiGibson: Sections like this are just about supporting browser features. I wasn't sure whether to go through and delete them all. Maybe it's not super important either way?

amarzavery:
Lgtm. Can't approve from cell phone.

lib/policies/serializationPolicy.ts
```diff
@@ -30,50 +31,45 @@ export class SerializationPolicy extends BaseRequestPolicy {
   }
 
   public async sendRequest(request: WebResource): Promise<HttpOperationResponse> {
```
daschult: If you're removing the await call, then you should make this function non-async.

lib/axiosHttpClient.ts
daschult: Does this mean we don't support upload and download progress reporting in node.js now?

RikkiGibson: These handlers never had an effect in node.js, which I guess is why I was inclined to remove them now that we only expect to use this module in node.js environments. We have to instead ask our node.js users to wrap their request body stream to get upload progress notifications.

lib/policies/serializationPolicy.ts
RikkiGibson: Good catch

PR #149, opened 2018-06-25
Move successful response deserialization to runtime

daschult:
Resolves https://github.com/Azure/ms-rest-js/issues/97.

This also removes the HttpRequest parameter from the sendOperationRequest() method.

### DIFF
--- a/lib/policies/serializationPolicy.ts
+++ b/lib/policies/serializationPolicy.ts
@@ -145,7 +145,7 @@ export class SerializationPolicy extends BaseRequestPolicy {
             valueToDeserialize = typeof valueToDeserialize === "object" ? valueToDeserialize[responseSpec.bodyMapper.xmlElementName!] : [];
           }
           try {
-          response.parsedBody = operationSpec.serializer.deserialize(responseSpec.bodyMapper, valueToDeserialize, "operationRes.parsedBody");
+            response.parsedBody = operationSpec.serializer.deserialize(responseSpec.bodyMapper, valueToDeserialize, "operationRes.parsedBody");
           } catch (error) {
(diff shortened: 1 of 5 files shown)


lib/policies/serializationPolicy.ts
```diff
@@ -138,8 +138,25 @@ export class SerializationPolicy extends BaseRequestPolicy {
...
         }
-      } else {
+      } else if (responseSpec) {
+        if (responseSpec.bodyMapper) {
+          let valueToDeserialize: any = response.parsedBody;
+          if (operationSpec.isXML && responseSpec.bodyMapper.type.name === MapperType.Sequence) {
+            valueToDeserialize = typeof valueToDeserialize === "object" ? valueToDeserialize[responseSpec.bodyMapper.xmlElementName!] : [];
+          }
+          try {
+          response.parsedBody = operationSpec.serializer.deserialize(responseSpec.bodyMapper, valueToDeserialize, "operationRes.parsedBody");
```
RikkiGibson: reindent

daschult: Fixed in https://github.com/Azure/ms-rest-js/pull/149/commits/b36d99cdac1d84aefe19cd003aaca338aae97810.

lib/serviceClient.ts
```diff
@@ -270,6 +273,18 @@ export class ServiceClient {
         }
       }
 
+      if (operationSpec.responses) {
+        let rawResponse = false;
```
RikkiGibson: this can maybe be more compact with `httpResponse.rawResponse = Object.values(operationSpec.responses).some(spec => ...)` or similar, but it's not super important

daschult: I'll leave it this way for now, but sure, we can change it later.

PR #153, opened 2018-06-26
Make OperationSpec.baseUrl an optional property for moving OperationSpecs into module/file scope

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/111.

These are the necessary changes in the runtime to move OperationSpecs out of the functions that use them.

### DIFF
--- a/lib/serviceClient.ts
+++ b/lib/serviceClient.ts
@@ -76,6 +76,12 @@ export interface ServiceClientOptions {
  */
 export class ServiceClient {
   /**
+   * If specified, this is the base URI that requests will be made against for this ServiceClient.
+   * If it is not specified, then all OperationSpecs must contain a baseUrl property.
+   */
+  protected baseUri?: string;
+
+  /**
    * The string to be appended to the User-Agent header while sending the request.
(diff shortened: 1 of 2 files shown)


lib/serviceClient.ts
```diff
@@ -177,6 +177,13 @@ export class ServiceClient {
 
     let result: Promise<HttpOperationResponse>;
     try {
+      if (operationSpec.baseUrl == undefined) {
+        operationSpec.baseUrl = (this as any).baseUri;
```
RikkiGibson: any reason baseUri can't also be an optional property on ServiceClient?

daschult: That's a good suggestion. I'll do that now.

amarzavery: Speaking of base uris have we ensured that we process x-ms-parameterized-host extension correctly in the new operation spec way of building the request?

daschult: Yes, we definitely do.

daschult: Fixed in https://github.com/Azure/ms-rest-js/pull/153/commits/8be6cd370d3075c9c33c878912d2b5d94235acd3.

PR #159, opened 2018-06-28
Support x-nullable

### LINKED ISSUE #135: Support XNullable in TypeScript SDK
Ensure that if a property is required then it must be not undefined. Null is a valid value for a required property if it is nullable.

RikkiGibson:
Fixes #135

### DIFF
--- a/lib/serializer.ts
+++ b/lib/serializer.ts
@@ -82,15 +82,30 @@ export class Serializer {
       payload = [];
     }
 
-    if (object == undefined) {
-      // Throw if required and object is null or undefined
-      if (mapper.required && !mapper.isConstant) {
-        throw new Error(`${objectName} cannot be null or undefined.`);
-      }
-      // Set Defaults
-      if (mapper.defaultValue != undefined || mapper.isConstant) {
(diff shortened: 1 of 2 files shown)


lib/serializer.ts
```diff
@@ -578,7 +584,7 @@ function deserializeCompositeType(serializer: Serializer, mapper: CompositeMappe
       if (Array.isArray(responseBody[key]) && modelProps[key].serializedName === "") {
         propertyInstance = responseBody[key];
         instance = serializer.deserialize(propertyMapper, propertyInstance, propertyObjectName);
-      } else if (propertyInstance !== null && propertyInstance !== undefined) {
+      } else if (propertyInstance !== undefined) {
```
daschult: `propertyInstance != undefined`?

RikkiGibson: I think I had a particular case here where I wanted to include a null value on the `instance` when deserializing, but drop anything that came back `undefined`.

One place where this would happen is if the property didn't exist on the parsed object tree, but we descended into `responseBody[key]` so we could check if it meets the required/nullable criteria for that property's spec.

PR #164, opened 2018-07-06
Allow withCredentials: true

### LINKED ISSUE #162: Introduce a way to enable authenticated CORS requests
Currently there appears to be no way to send authenticated CORS requests from browser. A naive quick fix would be to enable withCredentials in xhrHttpClient (see PR #161)

RikkiGibson:
Closes #162
Continuation of #161

- `withCredentials` will be provided in the `ServiceClientOptions`-- we assume that this is held in common for all requests made with a client, just like the credentials itself. We maybe could do some tricky inference to decide "oh, we have credentials, and this request is cross origin, so let's try and do the right thing"--but this seems good enough.
- Removing the unused `requestInit` property which is a breaking change

### DIFF
--- a/lib/serviceClient.ts
+++ b/lib/serviceClient.ts
@@ -31,11 +31,6 @@ import { RequestPrepareOptions, WebResource } from "./webResource";
  */
 export interface ServiceClientOptions {
   /**
-   * @property {RequestInit} [requestOptions] The request options. Detailed info can be found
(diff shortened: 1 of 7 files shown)


RikkiGibson:
Thanks for your contribution @R0nd!

(just noticed our changelog is a bit behind--I'll want to maybe go back over the commit history and fill that in..)

PR #171, opened 2018-07-11
Get client properties from ServiceClient instead of OperationArguments

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/189.

### DIFF
--- a/lib/serviceClient.ts
+++ b/lib/serviceClient.ts
@@ -192,7 +192,7 @@ export class ServiceClient {
       }
       if (operationSpec.urlParameters && operationSpec.urlParameters.length > 0) {
         for (const urlParameter of operationSpec.urlParameters) {
-          let urlParameterValue: string = getOperationArgumentValueFromParameter(operationArguments, urlParameter, operationSpec.serializer);
+          let urlParameterValue: string = getOperationArgumentValueFromParameter(this, operationArguments, urlParameter, operationSpec.serializer);
           urlParameterValue = operationSpec.serializer.serialize(urlParameter.mapper, urlParameterValue, getPathStringFromParameter(urlParameter));
           if (!urlParameter.skipEncoding) {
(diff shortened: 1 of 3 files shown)


lib/serviceClient.ts
```diff
@@ -384,25 +384,23 @@ export function getPropertyParent(parent: PropertyParent, propertyPath: string[]
...
   if (operationArguments.arguments) {
     if (Array.isArray(parameterPath)) {
       if (parameterPath.length > 0) {
-        value = operationArguments.arguments;
-        for (const parameterPathPart of parameterPath) {
-          value = value[parameterPathPart];
-          if (value == undefined) {
-            break;
-          }
+        let propertySearchResult: PropertySearchResult = getPropertyFromParameterPath(operationArguments.arguments, parameterPath);
```
RikkiGibson: This feels like a nice place for `const { propertyFound, propertyValue } = ...`

lib/serviceClient.ts
```diff
@@ -427,3 +425,27 @@ function getOperationArgumentValueFromParameterPath(operationArguments: Operatio
...
+  propertyValue?: any;
+  propertyFound: boolean;
+}
+
+function getPropertyFromParameterPath(parent: { [parameterName: string]: any }, parameterPath: string[]): PropertySearchResult {
+  const result: PropertySearchResult = { propertyFound: false };
+  let i = 0;
+  for (; i < parameterPath.length; ++i) {
+    const parameterPathPart: string = parameterPath[i];
+    // Make sure to check inherited properties too, so don't use hasOwnProperty().
```
RikkiGibson: `in parent` means that the parent contains a property with specifically that name?

daschult: Yup. That's what I want.

lib/serviceClient.ts
daschult: Can't do that here, because I may reassign propertySeachResult in two lines.

PR #172, opened 2018-07-14
Add responseToBody method

RikkiGibson:
Resolves https://github.com/Azure/autorest.typescript/issues/192.

responseToBody enables a much more compact implementation of the body overloads in generated code.

Before:
```ts
  getValid(options?: msRest.RequestOptionsBase, callback?: msRest.ServiceCallback<Models.Basic>): any {
    if (!callback && typeof options === 'function') {
      callback = options;
      options = undefined;
    }
    let cb = callback as msRest.ServiceCallback<Models.Basic>;
    if (!callback) {
      return this.getValidWithHttpOperationResponse(options).then((operationRes: msRest.HttpOperationResponse) => {
        return Promise.resolve(operationRes.parsedBody as Models.Basic);
      }).catch((err: Error) => {
        return Promise.reject(err);
      });
    } else {
      msRest.promiseToCallback(this.getValidWithHttpOperationResponse(options))((err: Error, data: msRest.HttpOperationResponse) => {
        if (err) {
          return cb(err);
        }
        let result = data.parsedBody as Models.Basic;
        return cb(err, result, data.request, data);
      });
    }
  }
```

After:
```ts
  getValid(options?: msRest.RequestOptionsBase, callback?: msRest.ServiceCallback<Models.Basic>): any {
    return msRest.responseToBody(this.getValidWithHttpOperationResponse.bind(this), options, callback);
  }
```

### DIFF
--- a/lib/util/utils.ts
+++ b/lib/util/utils.ts
@@ -14,6 +14,23 @@ import { Constants } from "./constants";
 export const isNode = typeof navigator === "undefined" && typeof process !== "undefined";
 
 /**
+ * Converts a WithHttpOperationResponse method to a method returning only the body.
+ * Supports an optional callback as the last argument.
+ * @param httpOperationResponseMethod
+ * @param args
+ */
+export function responseToBody(httpOperationResponseMethod: Function, ...args: any[]) {
+  const callback = args[args.length-1];
+  if (typeof callback === "function") {
+    httpOperationResponseMethod(...args.slice(0, args.length-1))
+      .then((res: HttpOperationResponse) => callback(null, res.parsedBody, res.request, res))
+      .catch((err: any) => callback(err));
+  } else {
+    return httpOperationResponseMethod(...args).then((res: HttpOperationResponse) => res.parsedBody);
+  }
+}
+
+/**
  * Checks if a parsed URL is HTTPS
  *
  * @param {object} urlToCheck The url to check
@@ -184,6 +201,7 @@ export interface ServiceCallback<TResult> {
  * Converts a Promise to a callback.
  * @param {Promise<any>} promise The Promise to be converted to a callback
  * @returns {Function} A function that takes the callback (cb: Function): void
+ * @deprecated generated code should instead depend on responseToBody
  */
 export function promiseToCallback(promise: Promise<any>): Function {
   if (typeof promise.then !== "function") {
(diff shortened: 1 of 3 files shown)


lib/util/utils.ts
```diff
@@ -184,6 +201,7 @@ export interface ServiceCallback<TResult> {
  * Converts a Promise to a callback.
  * @param {Promise<any>} promise The Promise to be converted to a callback
  * @returns {Function} A function that takes the callback (cb: Function): void
+ * @deprecated generated code should instead depend on responseToBody
```
RikkiGibson: Holding back on actual removal because I know ms-rest-nodeauth and possibly others depend on this function

RikkiGibson:
I suspect this might not handle the case where options are not provided but a callback is provided, so I'll probably want to add test coverage in ms-rest-js itself.

lib/util/utils.ts
```diff
@@ -13,6 +13,29 @@ import { Constants } from "./constants";
...
+ * @param httpOperationResponseMethod the method to call and apply a callback to
+ * @param args the arguments to the method, optionally including a trailing callback.
+ */
+export function responseToBody(httpOperationResponseMethod: (...args: any[]) => Promise<HttpOperationResponse>, ...args: any[]): Promise<any> | undefined {
+  // The generated code will always pass (options, callback) as the last two args.
+  // But `options` could actually be the callback for the sake of user convenience.
+  let callback = args.pop();
+  if (typeof callback !== "function" && typeof args[args.length - 1] === "function") {
+    callback = args.pop();
+  }
```
daschult: I need a little bit of explanation here. I know what should be happening (remove callback from the end of the argument list, if it is there). However, doesn't `pop()` remove the last argument from the list? What if no arguments are provided? What if the only argument is a value destined for the `httpOperationResponseMethod`? Won't `pop()` remove that argument so that `httpOperationResponseMethod` won't be provided the argument that it should be?
Also, I'm confused by line 25. If the last argument in the `args` array wasn't a callback function, but the second-to-last argument (in the original unmodified array) is, then we call `pop()` again? Doesn't your comment explain that there should never be an argument after the callback function?

test/shared/utilsTests.ts
```diff
@@ -0,0 +1,70 @@
...
+            return Promise.resolve(response);
+        };
+
+        const result = responseToBody(responseMethod, (err: any, body: any, request: WebResource, innerResponse: HttpOperationResponse) => {
+            should(err).equal(null);
+            body.foo.should.equal(42);
+            request.should.equal(response.request);
+            innerResponse.should.equal(response);
+            done();
+        }, undefined);
```
daschult: When do we ever call the callback from the second-to-last argument position?

PR #174, opened 2018-07-17
Timeout

### LINKED ISSUE #167: Expose HTTP request timeout
XMLHttpRequest and Axios both expose a `timeout` property which our users would like to make use of. This is a bit more convenient than e.g. an AbortController with a setTimeout that calls the `.abort()` after a certain amount of time (and also doesn't require including AbortController). @XiaoningLiu

RikkiGibson:
Fixes #167

### DIFF
--- a/lib/webResource.ts
+++ b/lib/webResource.ts
@@ -53,6 +53,7 @@ export class WebResource {
   query?: { [key: string]: any; };
   operationSpec?: OperationSpec;
   withCredentials: boolean;
+  timeout: number;
 
   abortSignal?: AbortSignalLike;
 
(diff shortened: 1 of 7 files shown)


lib/webResource.ts
```diff
@@ -68,21 +69,23 @@ export class WebResource {
     body?: any,
     query?: { [key: string]: any; },
     headers?: { [key: string]: any; } | HttpHeaders,
-    rawResponse = false,
```
RikkiGibson: Changed this because I don't trust the compiler to emit something that works in all environments

lib/xhrHttpClient.ts
```diff
@@ -127,4 +128,5 @@ function parseHeaders(xhr: XMLHttpRequest) {
 function rejectOnTerminalEvent(request: WebResource, xhr: XMLHttpRequest, reject: (err: any) => void) {
   xhr.addEventListener("error", ev => reject(new RestError(ev.message, "REQUEST_SEND_ERROR", undefined, request)));
   xhr.addEventListener("abort", () => reject(new RestError("The request was aborted", "REQUEST_ABORTED_ERROR", undefined, request)));
+  xhr.addEventListener("timeout", () => reject(new RestError(`timeout of ${xhr.timeout}ms exceeded`, "REQUEST_SEND_ERROR", undefined, request)));
```
RikkiGibson: Message designed to match the format of the axios timeout message

PR #177, opened 2018-07-25
Call progress callbacks in nodejs streams

### LINKED ISSUE #168: Consistent progress reporting
It's been requested that we make progress reporting feel consistent between node.js and browser by taking the user's progress handlers in nodejs and automatically wrapping their request/response body streams to call the progress handlers. @XiaoningLiu

Possible concerns:
- If the body is not a stream, we can't really report progress anyway in nodejs. Do we throw? Or warn? Or do nothing?

RikkiGibson:
Fixes #168

This strives for an as-uniform-as-possible experience between nodejs and browser. The progress callbacks will be called for both stream and buffer/string arguments--it's just (I assume) there will only be one progress event in the buffer case.

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -4,30 +4,26 @@
 import axios, { AxiosError, AxiosRequestConfig, AxiosResponse } from "axios";
 import { Transform, Readable } from "stream";
 import * as FormData from "form-data";
-import * as tough from "isomorphic-tough-cookie";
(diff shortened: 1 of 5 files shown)


RikkiGibson:
nodejs 6 requires I pass a Buffer instead of Uint8Array. Will fix.

amarzavery:
Do we need to support 6.x. 10.x will become LTS in this fall. In that respect 6.x will become ancient. I would recommend our min. support should be 8.x

PR #181, opened 2018-08-02
Add "sideEffects": false for webpack 4

RikkiGibson:
When bundling a sample I found this flag prevented webpack from including a redundant copy of the runtime in the bundle.


RikkiGibson:
Thanks to @TheLarkInn for the tip

PR #182, opened 2018-08-03
Typescript 3

### LINKED ISSUE #157: Fix usages of environment-specific globals
Particularly:

- [x] AbortSignal
- [ ] Blob

There may be some node-specific ones. Basically, someone writing a browser-only project should not get errors about nodejs types not being present, and someone writing a nodejs-only project should not get errors about browser types not being present.

RikkiGibson:
Fixes #157

### DIFF
--- a/.scripts/unit.js
+++ b/.scripts/unit.js
@@ -2,7 +2,7 @@ const { major } = require("semver");
 const { spawn, exec } = require("child_process");
 const { join } = require("path");
 
-const webpackDevServer = spawn(join(__dirname, "../node_modules/.bin/ts-node"), [join(__dirname, "../testserver")], { shell: true })
+const webpackDevServer = spawn(join(__dirname, "../node_modules/.bin/ts-node"), ["-T", join(__dirname, "../testserver")], { shell: true })
 function cleanupDevServer() {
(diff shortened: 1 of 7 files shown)


RikkiGibson:
I want VS Code to come out with a version that embeds TypeScript 3 before publishing this so that people don't get bogus errors related to the new `/// <reference lib="dom" />` syntax.

RikkiGibson:
VS Code has shipped a version with TypeScript 3 in the box, and since there's a breaking change in the lib.dom.d.ts I want to fix this up and get it merged.

RikkiGibson:
So, the TypeScript team has fixed the <reference lib="..." /> emit issue in a dev version. But I'm still leaning against it at this moment for 2 reasons:

1. People who have not updated to TypeScript 3 will get a hard-to-understand build error
2. Anyone who installs ms-rest-js in a node.js TypeScript project will suddenly get all the DOM APIs in the global scope--which won't exist at runtime, potentially causing confusion.

I've replaced the reference lib directive with an empty stub interface which gets augmented when lib.dom.d.ts is included, and since we only return values of this type, this seems adequate.

PR #187, opened 2018-08-13
Browser xml

RikkiGibson:
Drops isomorphic-xml2js and uses a simplified implementation with no runtime configuration supported. Saves 4 KB--now the overhead all users have to take on for XML support is just 2 KB.

Passes autorest.typescript tests.


RikkiGibson:
Looks like some string comparisons are broken in tests that expect a `<?` header element. This is low priority but would be nice to get in at some point.

RikkiGibson:
Seems like we're all green on this now, so I'll just merge and if problems arise for XML users we can look into it.

PR #190, opened 2018-08-15
Add validate flag to serializer

RikkiGibson:
Related to Azure/autorest.typescript#243


RikkiGibson:
Turns out this goes well beyond what autorest's `client-side-validation` is supposed to do.

https://github.com/Azure/autorest/blob/41c3ebf69e2903497ac32a430c27a972a03acb16/Samples/1d-common-settings/readme.md

RikkiGibson:
Some of the changes in here could still be used to reduce the size of the serializer a bit. Might be worth having a look at down the line.

PR #191, opened 2018-08-16
Add base HttpResponse

RikkiGibson:
Related to Azure/autorest.typescript#237

### DIFF
--- a/lib/httpOperationResponse.ts
+++ b/lib/httpOperationResponse.ts
@@ -5,10 +5,9 @@ import { WebResource } from "./webResource";
 import { HttpHeaders } from "./httpHeaders";
 
 /**
- * Wrapper object for http request and response. Deserialized object is stored in
- * the `parsedBody` property when the response body is received in JSON or XML.
+ * The properties on an HTTP response which will always be present.
  */
-export interface HttpOperationResponse<TBody = any, THeaders = any> {
+export interface BaseHttpOperationResponse {
   /**
    * The raw request
    */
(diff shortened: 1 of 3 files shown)


lib/httpOperationResponse.ts
```diff
@@ -5,10 +5,9 @@ import { WebResource } from "./webResource";
 import { HttpHeaders } from "./httpHeaders";
 
 /**
- * Wrapper object for http request and response. Deserialized object is stored in
- * the `parsedBody` property when the response body is received in JSON or XML.
+ * The properties on an HTTP response which will always be present.
  */
-export interface HttpOperationResponse<TBody = any, THeaders = any> {
+export interface BaseHttpOperationResponse {
```
daschult: We previously thought about renaming HttpOperationResponse to HttpResponse. Would this be a good time to do that, rather that adding the "Base" prefix?

RikkiGibson: So you think HttpResponse as the base, HttpOperationResponse as the "bucket of all the optionals"?

daschult: Yeah. Your design seems to say that HttpOperationResponse is being kept around just for backwards compatability and that BaseHttpOperationResponse/HttpResponse is what should be used/extended moving forward.

RikkiGibson:
BTW, I'm removing the type parameters in HttpOperationResponse, so I'm bumping the minor version

package.json
```diff
@@ -5,7 +5,7 @@
     "email": "azsdkteam@microsoft.com",
     "url": "https://github.com/Azure/ms-rest-js"
   },
-  "version": "0.18.0",
+  "version": "0.19.0",
```
RikkiGibson: Ahh, forgot to update the changelog again. I might just grind through the old versions to fill out the changelog so we can start making a habit of it.

PR #196, opened 2018-08-22
Extract createOperationRequest() from sendOperationRequest()

daschult:
This change is needed for the LRO rework I'm doing.


daschult:
On second thought, I guess I never ended up using this. Nevermind!

PR #199, opened 2018-08-23
Headers fix

### LINKED ISSUE #195: Deserialize response headers as case-insensitive
@XiaoningLiu has pointed out that some browsers use different casing conventions for header names and this prevents some response headers from deserializing properly.

HTTP header names are specified as case insensitive, so let's do the following to solve this:
- Convert header names on the response to lower case before deserializing
- Make serializedName on the response headers mapper always lower case (I think we already do this)

RikkiGibson:
Resolves #195

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -13,7 +13,7 @@ import { WebResource, HttpRequestBody } from "./webResource";
 
 const axiosClient = axios.create();
 // Workaround for https://github.com/axios/axios/issues/1158
-axiosClient.interceptors.request.use(config => ({ ...config, method: config.method && config.method.toUpperCase() }));
+axiosClient.interceptors.request.use(config => ({ ...config, method: config.method && config.method.toUpperCase() as any }));
 
 /**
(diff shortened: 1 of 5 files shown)


RikkiGibson:
We found that Firefox has a slightly different behavior around empty string headers.

I fixed xhrHttpClient to handle it and added a few tests to exercise headers parsing.

lib/axiosHttpClient.ts
```diff
@@ -13,7 +13,7 @@ import { WebResource, HttpRequestBody } from "./webResource";
 
 const axiosClient = axios.create();
 // Workaround for https://github.com/axios/axios/issues/1158
-axiosClient.interceptors.request.use(config => ({ ...config, method: config.method && config.method.toUpperCase() }));
+axiosClient.interceptors.request.use(config => ({ ...config, method: config.method && config.method.toUpperCase() as any }));
```
RikkiGibson: `npm run local` actually caused me to use my local version of axios, which brought in new, stricter type definitions that required this change.

PR #201, opened 2018-08-24
Changes needed for LRO support

daschult:
I recently rewrote the way we handle LROs in ms-rest-azure-js and these are the required changes in ms-rest-js to support that rewrite.
I renamed `WebResource.rawResponse` to `WebResource.streamResponseBody`. I also added `WebResource.shouldDeserialize` and `WebResource.operationResponseGetter`. These are for customizing the `deserializationPolicy` for a single request/response. They allow us to specify conditions that indicate that a response should or should not be deserialized (`shouldDeserialize`) and if we are deserializing, then which operationResponse in the operationSpec should be used for deserialization (`operationResponseGetter`).

### DIFF
--- a/lib/policies/deserializationPolicy.ts
+++ b/lib/policies/deserializationPolicy.ts
@@ -10,7 +10,6 @@ import * as utils from "../util/utils";
 import { parseXML } from "../util/xml";
 import { WebResource } from "../webResource";
 import { BaseRequestPolicy, RequestPolicy, RequestPolicyFactory, RequestPolicyOptions } from "./requestPolicy";
-import { isBoolean } from "util";
 
 /**
  * Create a new serialization RequestPolicyCreator that will serialized HTTP request bodies as they
@@ -58,7 +57,7 @@ function shouldDeserializeResponse(parsedResponse: HttpOperationResponse): boole
   let result: boolean;
   if (shouldDeserialize === undefined) {
     result = true;
-  } else if (isBoolean(shouldDeserialize)) {
+  } else if (typeof shouldDeserialize === "boolean") {
     result = shouldDeserialize;
   } else {
     result = shouldDeserialize(parsedResponse);
--- a/lib/policies/deserializationPolicy.ts
+++ b/lib/policies/deserializationPolicy.ts
@@ -3,13 +3,14 @@
 
 import { HttpOperationResponse } from "../httpOperationResponse";
 import { OperationResponse } from "../operationResponse";
-import { OperationSpec } from "../operationSpec";
+import { OperationSpec, isStreamOperation } from "../operationSpec";
 import { RestError } from "../restError";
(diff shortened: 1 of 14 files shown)


lib/policies/deserializationPolicy.ts
```diff
@@ -37,109 +38,133 @@ export class DeserializationPolicy extends BaseRequestPolicy {
...
-                }
-                if (parsedErrorResponse.message) {
-                  error.message = parsedErrorResponse.message;
-                }
-              } else {
-                let internalError: any = parsedErrorResponse;
-                if (parsedErrorResponse.error) {
-                  internalError = parsedErrorResponse.error;
-                }
+async function getOperationResponse(parsedResponse: HttpOperationResponse): Promise<undefined | OperationResponse> {
```
RikkiGibson: where's the async part of this?

daschult: Line 50. Is this one of those cases where it's better to not use async/await due to code size constraints?

RikkiGibson: Avoiding async/await in the non-test code is good to minimize size.

More importantly though I'm struggling to see why `operationResponseGetter` sometimes returns a Promise and other times doesn't. Is it because it sometimes needs to choose the response spec based the result of some async operation?

daschult: I just took another look and realized that we must've changed getting the response body to a synchronous operation/variable a while ago. I was still under the impression that was an async operation. I'll remove the Promise return option.

RikkiGibson: OK, it seems like that should hopefully simplify the types of several methods and variables, and make the await unnecessary.

daschult:
@RikkiGibson Any more thoughts on this PR?

lib/policies/deserializationPolicy.ts
```diff
@@ -3,13 +3,14 @@
...
 import { OperationResponse } from "../operationResponse";
-import { OperationSpec } from "../operationSpec";
+import { OperationSpec, isStreamOperation } from "../operationSpec";
 import { RestError } from "../restError";
 import { Mapper, MapperType } from "../serializer";
 import * as utils from "../util/utils";
 import { parseXML } from "../util/xml";
 import { WebResource } from "../webResource";
 import { BaseRequestPolicy, RequestPolicy, RequestPolicyFactory, RequestPolicyOptions } from "./requestPolicy";
+import { isBoolean } from "util";
```
RikkiGibson: This looks like a nodejs built-in which will cause webpack to bring in a polyfill. Would `typeof thing === "boolean"` be adequate?

daschult: Probably. I thought that was coming from our "../util/utils" file when VS Code recommended it. Thanks for finding that.

PR #205, opened 2018-08-28
Add RestResponse type and more runtime deserialization features

RikkiGibson:
- Adds RestResponse type (used in core runtime code, and used in generated code for responses which have no response body or headers)
- Converts blobBody from `() => Promise` to `Promise` just to get the final breaking changes out of the way in one go.

to-do:

- [x] fix tests
- [x] remove `msRest.responseToBody`
- [x] bump minor version


lib/httpOperationResponse.ts
```diff
@@ -68,3 +68,11 @@ export interface HttpOperationResponse extends HttpResponse {
    */
   readableStreamBody?: NodeJS.ReadableStream;
 }
+
+/**
+ * The flattened response to a REST call.
```
RikkiGibson: doc can be better

lib/policies/exponentialRetryPolicy.ts
```diff
@@ -86,8 +86,8 @@ export class ExponentialRetryPolicy extends BaseRequestPolicy {
  * @param {RetryData} retryData  The retry data.
  * @return {boolean} True if the operation qualifies for a retry; false otherwise.
  */
-function shouldRetry(policy: ExponentialRetryPolicy, statusCode: number, retryData: RetryData): boolean {
-  if ((statusCode < 500 && statusCode !== 408) || statusCode === 501 || statusCode === 505) {
+function shouldRetry(policy: ExponentialRetryPolicy, statusCode: number | undefined, retryData: RetryData): boolean {
```
RikkiGibson: A try can result in a transient failure s.t. there's no response object to even inspect. I'm pretty sure I've made this change before but somehow it didn't make it in.

lib/policies/exponentialRetryPolicy.ts
```diff
@@ -150,7 +150,9 @@ function retry(policy: ExponentialRetryPolicy, request: WebResource, response: H
     // If the operation failed in the end, return all errors instead of just the last one
     requestError = retryData.error;
     return Promise.reject(requestError);
-  } else {
+  } else if (response) {
     return Promise.resolve(response);
+  } else {
+    return Promise.reject();
```
RikkiGibson: needs to reject with an actual error value in the case that there was no response and also no error (either that or use `response!` on 154)

PR #214, opened 2018-09-05
Avoid split for undefined objects

abhishekgoenka:


daschult:
Thanks for contributing!

PR #219, opened 2018-09-10
Add additional properties support to serializer

RikkiGibson:


lib/serializer.ts
```diff
@@ -414,44 +413,47 @@ function serializeDictionaryType(serializer: Serializer, mapper: DictionaryMappe
...
+
+    const modelMapper = serializer.modelMappers[className];
+    if (!modelMapper) {
+      throw new Error(`mapper() cannot be null or undefined for model "${className}".`);
+    }
+    modelProps = modelMapper.type.modelProperties;
+    if (!modelProps) {
+      throw new Error(`modelProperties cannot be null or undefined in the ` +
+        `mapper "${JSON.stringify(modelMapper)}" of type "${className}" for object "${objectName}".`);
+    }
```
daschult: Does it make sense to assign `mapper.type.modelProperties = modelProps;` at this point so that we don't have to do the same checks the next time this method is called with that mapper?

RikkiGibson:
It looks like there's more polymorphic discriminator work to do. Will handle in a separate PR.

RikkiGibson:
Travis is refusing to signal completion to GitHub, so I'm turning off the branch protection in order to merge.

PR #225, opened 2018-09-12
Tolerate unknown/missing polymorphic discriminators

RikkiGibson:
There are test cases where the response contains an unknown or missing discriminator value. This change makes it so any failure to look up a polymorphic mapper results in us simply falling back to the mapper specified for the parameter or return type.

We're now at 52.1 KB.


RikkiGibson:
Need to populate the default discriminator value when [de]serializing.

PR #232, opened 2018-09-20
Es interop

### LINKED ISSUE #231: Enable esModuleInterop in tsconfig
Right now we don't have `esModuleInterop: true` in tsconfig and this allows us to sometimes produce code that's not semantically valid. This causes some tools like rollup that consume our esm outputs to fail.

e.g.

```js
import * as uuid from "uuid/v4";
uuid();
```

Let's fix these usages by enabling `esModuleInterop: true` as well as `allowSyntheticDefaultImports: true` so the usage changes to:

```js
import uuid from "uuid/v4";
uuid();
```

This is a preview blocker.
@bterlson

RikkiGibson:
Resolves #231 

@bterlson

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -3,7 +3,7 @@
 
 import axios, { AxiosError, AxiosRequestConfig, AxiosResponse } from "axios";
 import { Transform, Readable } from "stream";
-import * as FormData from "form-data";
+import FormData from "form-data";
 import * as tough from "tough-cookie";
 import { HttpClient } from "./httpClient";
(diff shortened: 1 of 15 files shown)


bterlson:
👍👍

RikkiGibson:
Pretty mystified-- tests run great locally but I get `webpackDevMiddleware is not a function` in CI.

RikkiGibson:
Should not have been mystified :P

RikkiGibson:
Will want to keep an eye on usages like `import * as tough from "tough-cookie"` in case that turns out to break rollup or something.

PR #237, opened 2018-10-01
Add support for credentials to be a request policy

kpajdzik:


lib/serviceClient.ts
```diff
@@ -358,15 +358,20 @@ export function serializeRequestBody(serviceClient: ServiceClient, httpRequest:
...
   const factories: RequestPolicyFactory[] = [];
 
   if (options.generateClientRequestIdHeader) {
     factories.push(generateClientRequestIdPolicy(options.clientRequestIdHeaderName));
   }
 
   if (credentials) {
-    factories.push(signingPolicy(credentials));
+    const crendentialsPolicyFactory = credentials as RequestPolicyFactory;
+    if (!crendentialsPolicyFactory || !crendentialsPolicyFactory.create) {
```
daschult: `!credentialsPolicyFactory` will always be false, because you just checked `credentials` a couple of lines above. In C# the `as` keyword will return null if the value is not a `RequestPolicyFactory`, but in TypeScript it can be considered a plain-old C-style cast.

daschult: Please remove `!credentialsPolicyFactory`.

RikkiGibson: The nicest thing might be to replace with a **user defined type guard**
```ts
function isRequestPolicyFactory(obj: any): obj is RequestPolicyFactory {
  return typeof obj.create === "function";
}
```

lib/serviceClient.ts
```diff
@@ -358,15 +358,23 @@ export function serializeRequestBody(serviceClient: ServiceClient, httpRequest:
...
   const factories: RequestPolicyFactory[] = [];
 
   if (options.generateClientRequestIdHeader) {
     factories.push(generateClientRequestIdPolicy(options.clientRequestIdHeaderName));
   }
 
   if (credentials) {
-    factories.push(signingPolicy(credentials));
+    if (isRequestPolicyFactory(credentials)) {
+      factories.push(credentials as RequestPolicyFactory);
```
RikkiGibson: is `as` still needed in this code?

PR #243, opened 2018-10-04
Rollup

### LINKED ISSUE #235: Rollup browser packaging
Making an issue for this as I have to park it due to bugs and tooling hangups. Will keep the WIP in the `rollup` branch.

RikkiGibson:
Resolves #235

- create node and browser rollup bundles
- ~~rename browser global to `Microsoft.Rest`~~
  - ~~ms-rest-azure-js will use `Azure.Rest`~~
  - ~~This doesn't incur a meaningful size penalty in the generated SDKs~~
- move .d.ts files to be inline with sources to align with conventions
- add declarationMap in tsconfig
- keep tsconfig.json with module: "commonjs" so that ts-node and similar tools will "just work"
- add ThirdPartyNotices because we're embedding uuid/v4 (MIT license) in the browser bundle (it's ~400 bytes, believe it's worth it over making end users put uuid in a script tag)
- keeping current webpack test harness-- would be ideal to start depending on the actually produced rollup bundle for tests though.

The browser script tag bundle is ~50 KB down from ~55 KB 👍


package.json
```diff
@@ -84,13 +87,12 @@
...
     "build:scripts": "tsc -p ./.scripts/",
-    "build:node": "tsc",
-    "build:es": "tsc -p tsconfig.es.json",
-    "build:browser": "webpack && node node_modules/uglify-es/bin/uglifyjs --source-map -c -m -o msRestBundle.min.js msRestBundle.js",
-    "watch:node": "tsc -w",
-    "watch:browser": "webpack -w",
+    "build:lib": "run-s build:tsc build:rollup build:minify-browser",
+    "build:tsc": "tsc -p tsconfig.es.json",
+    "build:rollup": "rollup -c rollup.config.js",
+    "build:minify-browser": "uglifyjs -c -m --comments --source-map \"content='./dist/msRest.browser.js.map'\" -o ./dist/msRest.browser.min.js ./dist/msRest.browser.js",
```
daschult: Changing from msRestBundle.js to msRest.browser.js is a breaking change, right? Shouldn't we bump the major version for this?

PR #251, opened 2018-10-17
Ignore XML headers when deserializing XML elements with headers and text

daschult:
This PR is all about this XML scenario:
```xml
<element attribute="value">text</element>
```
OpenAPI doesn't handle this scenario at all, so in your schema you have to either specify the attributes or the text. If you specify the attributes, then our generator would create a CompositeMapper. If you specify the text property (see the tests for an example), then our serializer was seeing a "StringMapper" but was given a composite object. As such, the serializer would just pass the composite object through unchanged.
This PR changes that so that the body gets passed through.

### DIFF
--- a/lib/serializer.ts
+++ b/lib/serializer.ts
@@ -167,11 +167,6 @@ export class Serializer {
 
     if (mapperType.match(/^Composite$/ig) !== null) {
       payload = deserializeCompositeType(this, mapper as CompositeMapper, responseBody, objectName);
-    } else if (mapperType.match(/^Number$/ig) !== null) {
-      payload = parseFloat(responseBody);
-      if (isNaN(payload)) {
-        payload = responseBody;
-      }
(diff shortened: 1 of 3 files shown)


lib/serializer.ts
```diff
@@ -165,35 +165,48 @@ export class Serializer {
...
+         * both header ("$") and body ("_") properties, then just reduce the responseBody value to
+         * the body ("_") property.
+         */
+        if (responseBody["$"] != undefined && responseBody["_"] != undefined) {
+          responseBody = responseBody["_"];
+        }
+      }
+
+      if (mapperType.match(/^Boolean$/ig) !== null) {
+        if (responseBody === "true") {
```
kpajdzik: Can that be simplified to:?
```suggestion
if (responseBody) {
  payload = responseBody;
}
```

Issue #228, opened 2018-09-12
Figure out a mechanism to add telemetry information for browser clients


amarzavery:
Currently we put package information in the user-agent header  for clients that run in the node.js runtime. However, in browser land, one cannot touch the user-agent header. It is forbidden to do so.

Hence we need to come up with a different mechanism to inject package name and version for telemetry.

/cc @RikkiGibson @kirthik  @JasonRShaver @salameer

kpajdzik:
> Had a short syncup with Jason on Friday. We have agreed to use x-ms-command-name  and x-ms-parameter-name request headers for sending telemetry for browser clients. We will be using the standard user-agent header for node.js clients. In this way it is easy to distinguish usage of browser and node.js sdks.

seiffs:
Could you please advise how to disable sending the telemetry headers? After upgrade from ms-rest-js to @azure/ms-rest-js there is now this new x-ms-command-name header and I can't find any way how to remove it. Unfortunatelly our API GW has very strict header rules, so all the request now fail on the preflight OPTIONS request because of this extra header.

kpajdzik:
@seiffs You need to replace default UserAgent policy with an instance created by you ([code](https://github.com/Azure/ms-rest-js/blob/master/lib/policies/userAgentPolicy.ts#L36)) or remove it from already created default policies.

PR #252, opened 2018-10-25
Add browser telemetry

kpajdzik:
Closes https://github.com/Azure/ms-rest-js/issues/228


lib/serviceClient.ts
```diff
@@ -71,6 +70,10 @@ export interface ServiceClientOptions {
    * The content-types that will be associated with JSON or XML serialization.
    */
   deserializationContentTypes?: DeserializationContentTypes;
+  /**
+   * The string to be set to the telemetry header while sending the request.
+   */
+  overriddenUserAgent?: string;
```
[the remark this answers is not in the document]
daschult: I still say this should just be `userAgent`. From the customer's perspective, they're not overriding anything. They're just providing a user agent that they want us to use.

lib/serviceClient.ts
```diff
@@ -382,7 +361,7 @@ function createDefaultRequestPolicyFactories(credentials: ServiceClientCredentia
   }
 
   if (utils.isNode) {
```
daschult: Remove this check now that our UserAgentPolicy works with both Node.js and browsers.

lib/serviceClient.ts
```diff
@@ -12,7 +12,7 @@ import { isStreamOperation, OperationSpec } from "./operationSpec";
 import { deserializationPolicy, DeserializationContentTypes } from "./policies/deserializationPolicy";
 import { exponentialRetryPolicy } from "./policies/exponentialRetryPolicy";
 import { generateClientRequestIdPolicy } from "./policies/generateClientRequestIdPolicy";
-import { msRestUserAgentPolicy } from "./policies/msRestUserAgentPolicy";
+import { userAgentPolicy } from "./policies/telemetry/userAgentPolicyFactory";
```
daschult: Can you figure out a way to make this import the same as the other policy imports instead of having the `Factory` suffix?

package.json
```diff
@@ -41,7 +41,7 @@
     "ThirdPartyNotices.txt"
   ],
   "browser": {
-    "./es/lib/policies/msRestUserAgentPolicy.js": "./es/lib/policies/msRestUserAgentPolicy.stub.js",
+    "./es/lib/policies/telemetry/msRestUserAgentPolicy.js": "./es/lib/policies/telemetry/msRestUserAgentPolicy.browser.js",
```
amarzavery: @kpajdzik - Was there a reason to add telemetry in the file path over here? It doesn't match the relative locations of those files. There is no telemetry folder inside `./lib/policies/` ?

Reference https://github.com/Azure/ms-rest-js/issues/281

kpajdzik: I believe that's a result of moving the files into and out of the "telemetry" directory as a result of code review.

amarzavery: I see. I think this is the root cause of the import os problem. Can you please fix this and publish a new version to latest tag?

PR #258, opened 2018-11-05
Fix deserializing nested composite types with no polymorphic discriminator specified

kpajdzik:


test/shared/serializationTests.ts
```diff
@@ -1086,5 +1086,215 @@ describe("msrest", function () {
...
+          sibling: { fishtype: "shark", age: 15 }
+        }, "");
+
+        assert.strictEqual("shark", result.fishtype);
+        assert.strictEqual(10, result.age);
+        assert.strictEqual("shark", result.sibling.fishtype);
+        assert.strictEqual(15, result.sibling.age);
+      });
+
+      it("should be deserialized properly when polymorphicDiscriminator specified in the parent", function () {
```
daschult: `not specified in the parent`?

PR #259, opened 2018-11-06
Fix the issue of "Initializers are not allowed in ambient contexts" for 3.0.x version of TS

amarzavery:


amarzavery:
@bterlson - Please review

PR #262, opened 2018-11-07
 Fix (de)serializing polymorphic elements in an array

kpajdzik:


lib/serializer.ts
```diff
@@ -669,9 +669,10 @@ function getPolymorphicMapper(serializer: Serializer, mapper: CompositeMapper, o
     if (discriminatorName != undefined) {
       const discriminatorValue = object[discriminatorName];
       if (discriminatorValue != undefined) {
-        const indexDiscriminator = discriminatorValue === mapper.type.uberParent
+        const parent = mapper.type.uberParent || mapper.type.className;
```
daschult: Please rename `parent` to `typeName`. `uberParent` and `className` may actually point to the current type.

lib/serializer.ts
```diff
@@ -683,7 +684,9 @@ function getPolymorphicMapper(serializer: Serializer, mapper: CompositeMapper, o
 }
 
 function getPolymorphicDiscriminatorRecursively(serializer: Serializer, mapper: CompositeMapper): PolymorphicDiscriminator | undefined {
-  return  mapper.type.polymorphicDiscriminator || (mapper.type.uberParent && serializer.modelMappers[mapper.type.uberParent].type.polymorphicDiscriminator);
+  return  mapper.type.polymorphicDiscriminator
+    || (mapper.type.uberParent && serializer.modelMappers[mapper.type.uberParent] && serializer.modelMappers[mapper.type.uberParent].type.polymorphicDiscriminator)
+    || (mapper.type.className && serializer.modelMappers[mapper.type.className] && serializer.modelMappers[mapper.type.className].type.polymorphicDiscriminator);
```
daschult: Refactor these expressions to a function that takes a property name. (Remove the duplication, please).

test/shared/serializationTests.ts
```diff
@@ -57,7 +57,7 @@ describe("msrest", function () {
     });
 
     it("should correctly serialize a complex JSON object", function (done) {
-      let o1: any = {
+      const o1: any = {
```
sergey-shandar: nit: use `unknown` instead of `any` type.

PR #265, opened 2018-11-09
Rename NPM package to @azure/ms-rest-js

kpajdzik:


package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "ms-rest-js",
+  "name": "@azure/ms-rest-js",
   "author": {
     "name": "Microsoft Corporation",
     "email": "azsdkteam@microsoft.com",
     "url": "https://github.com/Azure/ms-rest-js"
   },
-  "version": "1.0.0",
+  "version": "1.1.0",
```
daschult: There's no need to bump the version number because we're publishing to a new package, right?

sergey-shandar: @daschult IMHO: It's no required but it's good to have higher version if we rename the package.

sergey-shandar:
Do we need Azure DevOps release pipeline for this?

daschult:
@kpajdzik and I talked about this for another minute, and even though its convenient to have all of our packages under `@azure`, when a customer generates their own clients that don't target Azure services, they'll have a dependency on `@azure/ms-rest-js`. I know it doesn't really matter, but do we care that they'll take a dependency on a package that has `azure` in the name, even though their client may have nothing to do with Azure?

sergey-shandar:
@daschult I'm Ok with `@azure`. Our main target is Azure and we develop in `Azure` GitHub org. We may have a note that the library can be used for other (non-Azure) services.

PR #269, opened 2018-11-10
Build matrix

sergey-shandar:


azure-pipelines.yml
```diff
@@ -1,47 +1,69 @@
...
+      Node10:
+        node.version: 10.x
+      Node11:
+        node.version: 11.x
+  steps:
+  - task: NodeTool@0
+    displayName: 'Install Node'
+    inputs:
+      versionSpec: $(node.version)
+  - task: Npm@1
```
kpajdzik: Should all these tasks have the same name?

Issue #253, opened 2018-10-31
Default exponential retry policy should handle 429 error


xingwu1:
The current retry policy (https://github.com/Azure/ms-rest-js/blob/master/lib/policies/exponentialRetryPolicy.ts#L91) didn't honor the 429 error code.

PR #277, opened 2018-11-19
Add throttle handling policy

kpajdzik:
Resolves https://github.com/Azure/ms-rest-js/issues/253.


lib/policies/throttlingRetryPolicy.ts
```diff
@@ -0,0 +1,71 @@
...
+      const delayInMs: number | undefined = ThrottlingRetryPolicy.parseRetryAfterHeader(retryAfterHeader);
+      if (delayInMs) {
+        return delay(delayInMs).then(_ => this._nextPolicy.sendRequest(httpRequest));
+      }
+    }
+
+    return httpResponse;
+  }
+
+  public static parseRetryAfterHeader(headerValue: string): number | undefined {
```
daschult: Instead of making this a static function, I'd recommend just moving it outside of the class altogether.

lib/policies/throttlingRetryPolicy.ts
```diff
@@ -0,0 +1,71 @@
...
+    const retryAfterInSeconds = Number(headerValue);
+    if (Number.isNaN(retryAfterInSeconds)) {
+      return ThrottlingRetryPolicy.parseDateRetryAfterHeader(headerValue);
+    } else {
+      return retryAfterInSeconds * 1000;
+    }
+  }
+
+  public static parseDateRetryAfterHeader(headerValue: string): number | undefined {
```
daschult: Same here. Just remove it from the class and make it a normal function.

lib/serviceClient.ts
```diff
@@ -365,6 +366,7 @@ function createDefaultRequestPolicyFactories(credentials: ServiceClientCredentia
   if (!options.noRetryPolicy) {
     factories.push(exponentialRetryPolicy());
     factories.push(systemErrorRetryPolicy());
+    factories.push(throttlingRetryPolicy());
```
daschult: I'd recommend making the throttling retry policy come after the deserialization policy. I know that you'll have to check !options.noRetryPolicy twice, but it makes sense that we shouldn't try to deserialize a 429 response.

test/shared/policies/throttlingRetryPolicyTests.ts
```diff
@@ -0,0 +1,133 @@
...
+        });
+
+        it("should return sleep interval value in milliseconds if parameter is a number", function (done) {
+            const retryAfter = ThrottlingRetryPolicy.parseRetryAfterHeader("1");
+            assert.equal(retryAfter, 1000);
+            done();
+        });
+
+        it("should return sleep interval value in milliseconds for full date format", function (done) {
+            const clock = sinon.useFakeTimers(new Date("Fri, 31 Dec 1999 23:00:00 GMT").getTime());
```
daschult: As long as this works and it doesn't have any vulnerabilities, This is fine.

PR #289, opened 2018-12-03
Add deserializing missing properties

kpajdzik:


lib/serializer.ts
```diff
@@ -554,6 +556,8 @@ function deserializeCompositeType(serializer: Serializer, mapper: CompositeMappe
         if (headerKey.startsWith(headerCollectionPrefix)) {
           dictionary[headerKey.substring(headerCollectionPrefix.length)] = serializer.deserialize((propertyMapper as DictionaryMapper).type.value, responseBody[headerKey], propertyObjectName);
         }
+
+        handledPropertyNames.push(headerKey);
```
kpajdzik: Adding handled properties in every path causes problems with nested properties with conflicting names.

lib/serializer.ts
```diff
@@ -618,6 +622,27 @@ function deserializeCompositeType(serializer: Serializer, mapper: CompositeMappe
         instance[responsePropName] = serializer.deserialize(additionalPropertiesMapper, responseBody[responsePropName], objectName + '["' + responsePropName + '"]');
       }
     }
+  } else {
+    const containsSerializedName = (modelProperties: {[propertyName: string]: Mapper}, serializedKey: string): boolean => {
+      for (const key of Object.keys(modelProperties)) {
```
sergey-shandar: nit: `Object.keys(modelProperties).find(…)`

test/shared/serializationTests.ts
```diff
@@ -1049,11 +1050,12 @@ describe("msrest", function () {
       deserializedSawshark.siblings.length.should.equal(1);
       deserializedSawshark.siblings[0].fishtype.should.equal("mutatedshark");
```
sergey-shandar: nit: Could we use deep strict equal here to compare two structures?

lib/serializer.ts
```diff
@@ -618,6 +622,27 @@ function deserializeCompositeType(serializer: Serializer, mapper: CompositeMappe
...
+        if (mapper.serializedName === serializedKey) {
+          return true;
+        }
+      }
+
+      return false;
+    };
+
+    const isSpecialProperty = (propertyName: string): boolean => {
+      return [ "$", "_" ].includes(propertyName);
```
sergey-shandar:
```ts
const isSpecialProperty = (propertyName: string): boolean => 
  [ "$", "_"].includes(propertyName)
```

lib/serializer.ts
```diff
@@ -618,6 +622,27 @@ function deserializeCompositeType(serializer: Serializer, mapper: CompositeMappe
         instance[responsePropName] = serializer.deserialize(additionalPropertiesMapper, responseBody[responsePropName], objectName + '["' + responsePropName + '"]');
       }
     }
+  } else {
+    const containsSerializedName = (modelProperties: {[propertyName: string]: Mapper}, serializedKey: string): boolean => {
+      for (const key of Object.keys(modelProperties)) {
+        const mapper: Mapper = modelProperties[key];
```
sergey-shandar: nit:
```ts
for (const [key, mapper] of Object.entries(modelProperties)) { 
   ...
```

test/shared/serializationTests.ts
kpajdzik: Unfortunately, no as there are some properties that are ignored in the comparison.

lib/serializer.ts
daschult: I don't agree with this suggestion. I think the foreach pattern is just fine here.

lib/serializer.ts
daschult: This is a good suggestion, though.

lib/serializer.ts
daschult: This is specific to the XML scenario, so maybe rename this to `isSpecialXMLProperty`. I'd also recommend changing it to use the normal function syntax (`function isSpecialXMLProperty(propertyName: string): boolean =>...`).

lib/serializer.ts
```diff
@@ -532,6 +532,20 @@ function serializeCompositeType(serializer: Serializer, mapper: CompositeMapper,
   return object;
 }
 
+function isSpecialXmlProperty(propertyName: string): boolean {
+  return [ "$", "_" ].includes(propertyName);
+}
+
+function containsSerializedName(modelProperties: {[propertyName: string]: Mapper}, serializedKey: string): boolean {
+  for (const [, mapper] of Object.entries(modelProperties)) {
```
sergey-shandar: nit: :-)
```ts
Object.values(modelProperties).some(mapper => mapper.serializedName === serializedKey)
```

lib/serializer.ts
```diff
@@ -618,6 +631,12 @@ function deserializeCompositeType(serializer: Serializer, mapper: CompositeMappe
         instance[responsePropName] = serializer.deserialize(additionalPropertiesMapper, responseBody[responsePropName], objectName + '["' + responsePropName + '"]');
       }
     }
+  } else {
+    for (const key of Object.keys(responseBody)) {
+      if (instance[key] === undefined && !handledPropertyNames.includes(key) && !containsSerializedName(modelProps, key) && !isSpecialXmlProperty(key)) {
```
daschult: You shouldn't need to check containsSerializedName() anymore, right?

PR #318, opened 2019-01-17
Update changelog to match published version number

lukaselmer:
Update changelog to match published version number on npm and package.json:

* https://www.npmjs.com/package/@azure/ms-rest-js?activeTab=versions
* https://github.com/Azure/ms-rest-js/blob/master/package.json#L8

image

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.3.0 - 2019-01-15
+## 1.5.0 - 2019-01-15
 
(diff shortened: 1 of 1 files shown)


kpajdzik:
Actually, this bug was in fact fixed in 1.3.0 but thank you for bringing to our attention that changelog is incomplete for 1.4.0 and 1.5.0 releases. Let me fix that in another PR.

kpajdzik:
Related: https://github.com/Azure/ms-rest-js/pull/319/files

PR #322, opened 2019-01-22
Add HTTP(S) over HTTP(S) proxy support

### LINKED ISSUE #320: HTTPS over HTTP proxy doesn't work
Related axios issue: https://github.com/axios/axios/issues/925

kpajdzik:
Fixes #320


lib/axiosHttpClient.ts
```diff
@@ -134,6 +131,22 @@ export class AxiosHttpClient implements HttpClient {
         timeout: httpRequest.timeout,
         proxy: convertToAxiosProxyConfig(httpRequest.proxySettings)
       };
+
+      let axiosClient: AxiosInstance;
+      if (httpRequest.proxySettings) {
+        const agent = tunnel.httpsOverHttp({
```
daschult: What are the security implications for using HTTPS over HTTP? Is the traffic not encrypted when it travels between the source and the proxy, but then is encrypted between the proxy and the destination?

lib/serviceClient.ts
```diff
@@ -410,7 +410,7 @@ function createDefaultRequestPolicyFactories(credentials: ServiceClientCredentia
 
   factories.push(deserializationPolicy(options.deserializationContentTypes));
 
-  if (options.proxySettings) {
+  if (options.proxySettings || (options.proxySettings = getDefaultProxySettings())) {
```
daschult: Assignment within an if-expression? Please pull that assignment out of the expression. ;)

PR #323, opened 2019-01-25
Fix tunnel types are missing

vladbarosan:
- type from tunnel is exposed but the types dependency is a dev dependency only
- use valid es6 module syntax for 'http' and 'https' packages

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -12,8 +12,8 @@ import { RestError } from "./restError";
 import { WebResource, HttpRequestBody } from "./webResource";
 import * as tunnel from "tunnel";
 import { ProxySettings } from "./serviceClient";
-import http from "http";
-import https from "https";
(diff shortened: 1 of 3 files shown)


vladbarosan:
Using latest autorest.typescript generator, the following error occurs on `npm pack`
```
> tsc && rollup -c rollup.config.js && npm run minify

node_modules/@azure/ms-rest-js/es/lib/axiosHttpClient.d.ts:6:25 - error TS7016: Could not find a declaration file for module 'tunnel'. 'C:/vladdb/devdiv/repos/azure/openapi-tools/sdk/azure-tooling-sdk-for-typescript/node_modules/tunnel/index.js' implicitly has an 'any' type.
  Try `npm install @types/tunnel` if it exists or add a new declaration (.d.ts) file containing `declare module 'tunnel';`
```

It seems that a type from tunnel is exported but the types package is only a dev dependency

PR #326, opened 2019-01-29
Fix 'ReferenceError: process is not defined' error in Angular

shigeyf:


kpajdzik:
@shigeyf, Could you share solution or steps to reproduce the issue?

Issue #310, opened 2019-01-14
Consider Standardizing on fetch for Rest Calls


justinwilaby:
The discussion at #8 addresses several issues that can be solved by standardizing on `fetch` for the browser and `node-fetch` for node. 
 
✓ Proxy support 
✓ FormData Support
✓ Standard API
✓ Zero dependencies for the web
✓ Reasonable number of dependencies for Node
✓ Isomorphic

RikkiGibson:
I wanted to note why we didn't end up using fetch as the person who was responsible for us moving away from it. (no longer working on the project)

- node-fetch didn't have abort support
- fetch implementations didn't have support for progress reporting (particularly useful for file uploads)

I'm sure these issues will be rectified eventually at which point the maintainers may revisit the question.

justinwilaby:
Looks like [Abort support](https://github.com/bitinn/node-fetch/blob/master/README.md#request-cancellation-with-abortsignal) was already added.

Since `node-fetch` can post using [file streams](https://github.com/bitinn/node-fetch#post-data-using-a-file-stream), it seems reasonable that upload progress can be achieved using the [data event](https://nodejs.org/api/stream.html#stream_event_data). This might also be available using the [web-api for ReadableStreams](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#Examples) as well.

ramya-rao-a:
`@azure/ms-rest-js` is currently in maintenance mode with no new features being added and taking only bug fixes. We are moving active development efforts in the azure-sdk-for-js repo on `@azure/core-http` that will power the next generation of the Azure packages which is using node-fetch and the standardized `@azure/abort-controller`. Therefore, closing this issue.

Issue #338, opened 2019-04-11
isNode returns false for electron apps


ramya-rao-a:
The current implementation for `isNode` is:

```javascript
export const isNode = typeof navigator === "undefined" && typeof process !== "undefined";
```

For electron apps this will return `false` and I believe we need it to return `true`.

Suggested edit:

```javascript
export const isNode = !!process && !!process.version && !!process.versions && !!process.versions.node;
```

ramya-rao-a:
cc @bterlson

kpajdzik:
`!!process.version` is this part necessary?

PR #342, opened 2019-05-06
Add Fetch HTTP client

kpajdzik:
- fixes https://github.com/Azure/ms-rest-js/issues/310
- fixes https://github.com/Azure/ms-rest-js/issues/8


lib/fetchHttpClient.ts
```diff
@@ -0,0 +1,210 @@
...
+import { Readable, Transform } from "stream";
+
+interface FetchError extends Error {
+    code: string | undefined;
+    errno: string | undefined;
+    type: string | undefined;
+}
+
+export class FetchHttpClient implements HttpClient {
+    private readonly cookieJar = new tough.CookieJar();
```
amarzavery: is tough-cookie isomorphic? I know, there were problems with it previously.

lib/fetchHttpClient.ts
```diff
@@ -0,0 +1,210 @@
...
+import { WebResource } from "./webResource";
+import { HttpOperationResponse } from "./httpOperationResponse";
+import { HttpHeaders } from "./httpHeaders";
+import { RestError } from "./restError";
+import { Readable, Transform } from "stream";
+
+interface FetchError extends Error {
+    code: string | undefined;
+    errno: string | undefined;
+    type: string | undefined;
```
amarzavery: Nit: Instead of having undefined, you could also use the question mark pattern. ex: `code?: string`

lib/fetchHttpClient.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+
+export class FetchHttpClient implements HttpClient {
+    private readonly cookieJar = new tough.CookieJar();
+
+    async sendRequest(httpRequest: WebResource): Promise<HttpOperationResponse> {
+        if (!httpRequest) {
```
amarzavery: We aren't checking the type of httpRequest. Someone can pass a boolean and the condition will still pass in JS.

`if (typeof httpRequest !== "object") {` // This will check for null, undefined cases and also does the type checking.

lib/fetchHttpClient.ts
```diff
@@ -0,0 +1,210 @@
...
+            throw new Error("httpRequest (WebResource) cannot be null or undefined and must be of type object.");
```
amarzavery: surround object with single quotes.

lib/fetchHttpClient.ts
```diff
@@ -0,0 +1,210 @@
...
+
+import { HttpClient } from "./httpClient";
+import { WebResource } from "./webResource";
+import { HttpOperationResponse } from "./httpOperationResponse";
+import { HttpHeaders } from "./httpHeaders";
+import { RestError } from "./restError";
+import { Readable, Transform } from "stream";
+
+interface FetchError extends Error {
+    code: string | undefined;
```
amarzavery: may be i am blind. Is the indentation 4 spaces? If so, then please change it to 2 spaces.

test/defaultHttpClientTests.ts
```diff
@@ -163,7 +163,7 @@ describe("defaultHttpClient", function () {
       download.notified.should.be.true;
     });
 
-    it("for blob or stream bodies", async function () {
+    it.skip("for blob or stream bodies", async function () {
```
amarzavery: reason to skip this test?

test/defaultHttpClientTests.ts
```diff
@@ -163,7 +163,7 @@ describe("defaultHttpClient", function () {
...
       let payload: HttpRequestBody;
       if (isNode) {
```
amarzavery: Did we change the logic to detect isNode? https://github.com/Azure/ms-rest-js/issues/338.
Please change it, since storage will be using electron and the runtime will create issues for them.

lib/fetchHttpClient.ts
kpajdzik: This is for Node-only right now

lib/fetchHttpClient.ts
kpajdzik: @daschult is `| undefined` kind of person so maybe we should discuss what's our official style ;)

test/defaultHttpClientTests.ts
kpajdzik: Reporting downloading/uploading is kind of hard with Fetch right now. The Storage Explorer team doesn't need it so I didn't spend too much time right now on it.

kpajdzik: If Fetch solves their problems I'll try to finish all the functionality.

test/defaultHttpClientTests.ts
kpajdzik: I'll create another PR for that

test/defaultHttpClientTests.ts
amarzavery: So here we are going backwards. I think this may affect storage js sdk. We will break them.

.vscode/launch.json
```diff
@@ -15,16 +15,6 @@
                 "--colors"
             ],
             "internalConsoleOptions": "openOnSessionStart"
-        },
-        {
```
RikkiGibson: Did this script stop working and that's why it's going away?

lib/fetchHttpClient.ts
RikkiGibson: I noticed you're using `code: string | undefined` instead of `code?: string`. is that because you want to require explicitly providing an `undefined`?

lib/fetchHttpClient.ts
```diff
@@ -0,0 +1,210 @@
...
+                request: httpRequest,
+                status: response.status
+            };
+
+            const onDownloadProgress = httpRequest.onDownloadProgress;
+            if (!onDownloadProgress) {
+                const bodyAsText = response.body ? await response.text() : undefined;
+                operationResponse.bodyAsText = bodyAsText;
+            } else {
+                const responseBody: ReadableStream<Uint8Array> | undefined = response.body ? response.body : undefined;
```
RikkiGibson: Does this ReadableStream come from fetch spec? Is it actually implemented in node now?

RikkiGibson: Maybe consider `response.body || undefined` if the goal is to normalize all falsy values to undefined

test/defaultHttpClientTests.ts
```diff
@@ -180,7 +180,7 @@ describe("defaultHttpClient", function () {
       const upload: Notified = { notified: false };
       const download: Notified = { notified: false };
 
-      const request = new WebResource("/fileupload", "POST", payload, undefined, undefined, true, undefined, undefined, 0,
+      const request = new WebResource("/fileupload", "POST", payload, undefined, { "Content-Length": 100 }, true, undefined, undefined, 0,
```
RikkiGibson: Is it necessary to provide Content-Length for a known-size request body?

test/defaultHttpClientTests.ts
RikkiGibson: I'm slightly surprised they don't require progress reporting..don't they have progress bars for uploads and downloads?

RikkiGibson: On the node side I thought this was already done by wrapping the body streams to provide reporting. As long as fetch lets you provide and receive a node-style stream for the request and response body it should be doable. And on the browser side it seems like you'll continue using XhrHttpClient(?) so that's not a concern.

lib/fetchHttpClient.ts
RikkiGibson: You probably already looked at this but:

```js
// in browser REPL
> typeof null
< "object"
```

lib/fetchHttpClient.ts
RikkiGibson: afaik they really are subtly different, one is an optional member while the other is required but you're allowed to pass `undefined` for it. So you should pick the one that has the behavior you want.

kpajdzik:
> lgtm barring progress reporting stuff

Thanks so much for taking time to review it, really appreciate it. 

Yes, progress reporting for streams needs to be added but I wanted to get early feedback from the team and Storage team :)

.vscode/launch.json
kpajdzik: We've moved away from the test server. It was frequently failing in CI environment and it was hard to reproduce the errors so we changed the test code to mocks.

kpajdzik: That was some time ago, removing to keep things up-to-date.

amarzavery:
@kpajdzik - Please pull from upstream master and also update the changelog

Issue #348, opened 2019-05-15
ChangeLog.md is not updated


XiaoningLiu:
Hi,

ChangeLog.md seems not fully updated with npm released versions, and doesn't include all release changes. Is this by purpose?

https://github.com/Azure/ms-rest-js/blob/master/Changelog.md
https://www.npmjs.com/package/@azure/ms-rest-js

amarzavery:
Fixed via #349 Here is the updated [changelog](https://github.com/Azure/ms-rest-js/blob/master/Changelog.md)

PR #349, opened 2019-05-16
Fixed #347 and #348

### LINKED ISSUE #347: reqOptions.headers.set is undefined
@RikkiGibson , https://github.com/Azure/ms-rest-js/blob/master/lib/policies/rpRegistrationPolicy.ts#L70 reqOptions.headers.set is undefined due to the empty object creation on line 61 with no members at all.

amarzavery:


lib/policies/rpRegistrationPolicy.ts
```diff
@@ -57,24 +57,17 @@ function registerIfNeeded(policy: RPRegistrationPolicy, request: WebResource, re
  * @returns {object} A new request object with desired headers.
  */
 function getRequestEssentials(originalRequest: WebResource, reuseUrlToo = false): any {
```
daschult: Change return type to be the real return type of the function.

lib/policies/rpRegistrationPolicy.ts
```diff
@@ -57,24 +57,17 @@ function registerIfNeeded(policy: RPRegistrationPolicy, request: WebResource, re
...
-  const reqOptions: any = {
-    headers: {}
-  };
+  const reqOptions = originalRequest.clone();
```
daschult: Add a type for `reqOptions`.

PR #352, opened 2019-06-02
Bumped version of axios to 0.19.0

amarzavery:


jeremymeng:
version in constants.ts also needs update.

PR #358, opened 2019-06-04
Set tranformResponse property of axios req. config to an identity function

amarzavery:
`axios` changed the way it treats properties of the request config in `0.19.0`. Previously we were setting `trasnformResponse` to `undefined`. This would indicate `axios` to not transform (`JSON.parse()`) the response body. In `0.19.0`, they are setting the default response transformer if transformResponse is set to `undefined`. This breaks our pasrsing logic where we are doing `JSON.parse()` on `operationResponse.bodyAsText`. Moreover, we are exposing the `bodyAsText` property in the generated clients. 
Not populating this property or setting the value of this property to a parsed JSON would be a breaking change for our users.
Hence we are setting the `transformResponse` property in the request config to an indentity function that returns the response body as-is.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 1.8.10 - 2019-06-05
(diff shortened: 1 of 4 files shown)


amarzavery:
https://github.com/Azure/ms-rest-azure-js/pull/119

PR #361, opened 2019-06-12
Added domainCredentials for authentication to EventGrid domains.

kalyanaj:
Added a new DomainCredentials class for make it easier to supply EventGrid domain credentials. Event Grid domain is a recently introduced concept/resource and applications can publish events to an EventGrid domain for which an api key needs to be provided, and this newly introduced DomainCredentials class makes it easy to do so.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.8.13 - 2019-06-11
+- Added DomainCredentials class for providing credentials to publish to an Azure EventGrid domain.
+
 ## 1.8.12 - 2019-06-07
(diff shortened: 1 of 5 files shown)


amarzavery:
Please update the version in package.json to 1.8.13 and update the Changelog.md with the changes for the new version. Also update the version to 1.8.13 over [here](https://github.com/Azure/ms-rest-js/blob/master/lib/util/constants.ts#L10).

kalyanaj:
Thanks @amarzavery for the review!  I have incorporated the changes you suggested, can you please take a look when you get a chance?

PR #362, opened 2019-06-18
Add keep-alive support

kpajdzik:


lib/nodeFetchHttpClient.ts
```diff
@@ -49,6 +51,16 @@ export class NodeFetchHttpClient extends FetchHttpClient {
       requestInit.agent = tunnel.agent;
     }
 
+    if (httpRequest.keepAlive === true) {
```
daschult: I'm guessing you did this because `keepAlive` can be undefined. Couldn't you also just do `if (httpRequest.keepAlive) {...}`?

kpajdzik: Correct, it can be undefined. I can default it to false but why litter httpRequest object and headers when there is no need for it.

PR #366, opened 2019-07-10
serialization updates

amarzavery:
- Added support to not send default values while sending the request.
- Added support to populate entities with it's default value if it is present in the mapper while deserializing the response.
- During deserialization, if the service does not provide the **_discriminator_** property then we set it. While setting the _**discriminator**_ property, we compare model property name and the `clientName` of the `polymorphicDiscriminator` instead of the `serializedName` of the `polymorphicDiscriminator`.
- Added tests for serializing and deserializing additional properties.

### DIFF
--- a/lib/serializer.ts
+++ b/lib/serializer.ts
@@ -85,7 +85,7 @@ export class Serializer {
       payload = [];
     }
 
-    if (object == undefined && (mapper.defaultValue != undefined || mapper.isConstant)) {
+    if (mapper.isConstant) {
       object = mapper.defaultValue;
     }
 
@@ -158,6 +158,9 @@ export class Serializer {
         // between the list being empty versus being missing,
         // so let's do the more user-friendly thing and return an empty list.
         responseBody = [];
+      }
+      if (mapper.defaultValue) {
+        responseBody = mapper.defaultValue;
       }
       return responseBody;
     }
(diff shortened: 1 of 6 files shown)


lib/serializer.ts
```diff
@@ -159,6 +159,9 @@ export class Serializer {
         // so let's do the more user-friendly thing and return an empty list.
         responseBody = [];
       }
+      if (mapper.defaultValue) {
```
RikkiGibson: are there any subtle bugs this could bring in related to falsy default values?

amarzavery: Thanks for pointing this out. Have changed this to specifically check for `undefined`.

amarzavery:
@bterlson, @daviwil - Please migrate the changes made in this PR to `@azure/core-http` as it fixes some important [de]serialization bugs.
- Was able to catch some bugs that were then fixed in the [second last commit](https://github.com/Azure/ms-rest-js/pull/366/commits/21541e9c06173b34e69ad20afa2c3315debee35c) in this PR, because we run `autorest.typescript` tests with the changes in this PR. This gives us confidence that we shouldn't be breaking anything by mistake. Please make sure to run `autorest.typescript` tests whenever changes are made to `@azure/core-http` once `autorest.typescript` moves to `@azure/core-http`.

daviwil:
Thanks!  I'll get `core-http` up to date with the latest changes of `ms-rest-js` soon.  I agree, it'd be good to run `autorest.typescript`'s tests when `core-http` changes, I'll talk to our engineering systems team about it.

amarzavery:
@daviwil - FYI, this [azure devops config file](https://github.com/Azure/ms-rest-js/blob/master/.devops/azure-pipelines-test-dependent-projects.yml) for running autorest.typescript tests against the changes made in the runtime can be helpful.

PR #368, opened 2019-07-16
Introduce IHttpHeaders to enable compatibility with @azure/core-http

daviwil:
This change introduces an `IHttpHeaders` interface (please give suggestions on the name) that will resolve a structural compatibility issue that prevents `ms-rest-nodeauth` from being used with SDK libraries generated with `@azure/core-http`.  The core issue is that the `HttpHeader` class uses a [`private` property](https://github.com/Azure/ms-rest-js/blob/master/lib/httpHeaders.ts#L35) which breaks structural compatibility with the same `HttpHeader` class in `@azure/core-http`.  The [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/classes.html#understanding-private) explains why:

> TypeScript is a structural type system. When we compare two different types, regardless of where they came from, if the types of all members are compatible, then we say the types themselves are compatible.
>
> However, **when comparing types that have private and protected members, we treat these types differently.** For two types to be considered compatible, if one of them has a private member, then the other must have a private member that originated in the same declaration. The same applies to protected members.

To ultimately resolve this issue, we'll have to do the following:

1. Ship a patch update in the 1.x series of `ms-rest-js` with this fix so that users of existing SDK libraries will pick it up
2. Ship a patch update in the 2.x series of `ms-rest-js` to ensure that anyone who might be using that range gets the fix
3. Ship an update to `@azure/core-http` with this fix

`ms-rest-nodeauth` itself should not need to be updated so long as its transitive dependency version of `ms-rest-js` gets updated to the fixed version.

**Question for @amarzavery**

I've built these changes on top of commit 4c755ad which was the last before the 2.x version series was introduced.  Can we create a `1.x` branch in this repo from that commit so that I can retarget this PR there?  We'll need to ship the patch update from that branch.

Once we settle on the details of this PR I'll send the equivalent changes for `master` of this repo and to `@azure/core-http`.

/cc @bterlson @ramya-rao-a

### DIFF
--- a/lib/msRest.ts
+++ b/lib/msRest.ts
@@ -4,7 +4,7 @@
 export { WebResource, HttpRequestBody, RequestPrepareOptions, HttpMethods, ParameterValue, RequestOptionsBase, TransferProgressEvent, AbortSignalLike } from "./webResource";
 export { DefaultHttpClient } from "./defaultHttpClient";
 export { HttpClient } from "./httpClient";
-export { HttpHeaders } from "./httpHeaders";
+export { IHttpHeaders, HttpHeaders } from "./httpHeaders";
 export { HttpOperationResponse, HttpResponse, RestResponse } from "./httpOperationResponse";
 export { HttpPipelineLogger } from "./httpPipelineLogger";
 export { HttpPipelineLogLevel } from "./httpPipelineLogLevel";
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.8.14 - 2019-07-17
+- Introduced `IHttpHeaders` interface to improve compatibility with `@azure/core-http` for users of `@azure/ms-rest-nodeauth`
+
 ## 1.8.13 - 2019-06-12
 - Added DomainCredentials class for providing credentials to publish to an Azure EventGrid domain.
 
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -6,7 +6,7 @@ import { Transform, Readable } from "stream";
 import FormData from "form-data";
 import * as tough from "tough-cookie";
 import { HttpClient } from "./httpClient";
-import { HttpHeaders } from "./httpHeaders";
+import { IHttpHeaders, HttpHeaders } from "./httpHeaders";
 import { HttpOperationResponse } from "./httpOperationResponse";
 import { RestError } from "./restError";
 import { WebResource, HttpRequestBody } from "./webResource";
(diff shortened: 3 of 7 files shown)


lib/msRest.ts
```diff
@@ -4,7 +4,7 @@
 export { WebResource, HttpRequestBody, RequestPrepareOptions, HttpMethods, ParameterValue, RequestOptionsBase, TransferProgressEvent, AbortSignalLike } from "./webResource";
 export { DefaultHttpClient } from "./defaultHttpClient";
 export { HttpClient } from "./httpClient";
-export { HttpHeaders } from "./httpHeaders";
+export { IHttpHeaders, HttpHeaders } from "./httpHeaders";
```
daviwil: I'm torn on whether this should actually be exported from the module's public interface as it seems unnecessary.  Thoughts?

amarzavery: I think, in TypeScript it is okay to have an interface and a class with the same name. So how about naming the interface HttpHeaders as well.

daviwil: That's a great idea!  Initial testing shows it might work, stay tuned...

amarzavery: Another thought, What if we name the interface HttpHeaders and the class as `HttpHeadersImpl`. This shouldn't be a breaking change in TypeScript since, TS focuses on the structure and not the name. AFAIK, you have simply added an interface that declares the methods and properties of what was the `HttpHeaders` **class**. If that is true, then renaming the class to `HttpHeadersImpl` should work. `HttpHeaders` is already exported. So everything should work business as usual.

daviwil: Looks like TypeScript can't disambiguate between the class an interface well enough when they have the same name.  For now I'll stick with `IHttpHeaders` unless someone has a better idea for its name

daviwil: (whoops, wrote my last comment without reloading to see @amarzavery's latest)

Already chatted with Amar offline about this, I think it'll be a breaking change to rename `HttpHeaders` to `HttpHeadersImpl` if anyone using `ms-rest-js` is creating a `HttpHeaders` instance.

bterlson:
I like this approach! I like HttpHeadersImpl better as a name for what that's worth.

daviwil:
Unfortunately renaming `HttpHeaders` to `HttpHeadersImpl` breaks the public interface of `ms-rest-js` since a user could be creating an instance of `HttpHeaders` (potentially for testing purposes, likely in one of our own libraries).  Retaining the `HttpHeaders` name for the class is key to not introducing a breaking change.

ramya-rao-a:
I know quite some work has gone into this, so, very sorry for throwing a wrench into the works here....  But, https://github.com/Azure/azure-sdk-for-js/issues/4108#issuecomment-512628908

daviwil:
As of https://github.com/Azure/azure-sdk-for-js/issues/4108#issuecomment-512948979, we've decided to take a different approach that doesn't require any changes to `ms-rest-js`.  Closing this PR for now!

PR #370, opened 2019-08-06
Update dev dependency yarn

mikeharder:

PR #371, opened 2019-08-08
Clone the v4x branch of autorest.typescript for testing purposes

daviwil:
This change fixes an issue that was discovered when PR #370 was created: this repo clones and builds the `autorest.typescript` repo during PR CI runs to verify that no unexpected breaking changes are introduced.  Since `autorest.typescript` v5 (on `master`) has made some API changes regarding credential types, we now need to explicitly clone the `v4x` branch of that repo so that the right interfaces are used.


amarzavery:
LGTM

PR #372, opened 2019-08-08
Update CODEOWNERS

daviwil:
Removes old maintainers from `CODEOWNERS` and replaces with the new.


.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
 # Default owners
-* @amarzavery @daschult @kpajdzik @sergey-shandar
+* @daviwil @ramya-rao-a
```
ramya-rao-a: Let's keep Sergey until we are more comfortable with the code here

Issue #374, opened 2019-10-11
Out of memory issue on making hundreds of requests


aby040:
**Package Version**: 1.8.13
- [x] **nodejs**
   - **nodejs version**: 10.16.0
   - **os name/version**: macOS Catalina
- [ ] **browser**
    - **name/version**: ____

**Describe the bug**
Out of memory issue on making multiple requests to fetch the list of `network-interfaces` using the `@azure/arm-network` package (version 14.0.0)

**To Reproduce**
1. Try to make hundreds of request to fetch the list of `network-interfaces` using the `@azure/arm-network` package (version 14.0.0).
2. Increase in the heap size can be observed after each request which is not getting cleared.

**Expected behavior**
There should not be any memory leak.

**Additional context**
Issue is not found on later versions of `@azure/arm-network` but there are other `@azure` packages which are still relying on versions `^1.8.1` of `@azure/ms-rest-js`

ramya-rao-a:
Fixed with #375

PR #375, opened 2019-10-11
Out of memory error on making multiple requests

aby040:
Fixes: #374 

Creating a separate axios instance and moving the interceptor code out of the sendRequest() method.

An interceptor was added to the static axios instance on each sendRequest() call which can lead to out of memory issue if hundreds of requests are made as the added interceptor is not ejected after the request is made.

### DIFF
--- a/.scripts/unit.ts
+++ b/.scripts/unit.ts
@@ -75,13 +75,13 @@ function stopTestServer(testServer: ServerProcess): void {
 
 function runNodeJsUnitTests(): number {
   console.log(`Running Node.js Unit Tests...`);
-  return executeSync(`nyc mocha`, repositoryRootFolderPath).status || 1;
+  return executeSync(`nyc mocha`, repositoryRootFolderPath).status;
 }
 
 function runBrowserUnitTests(): number {
   console.log(`Running Browser Unit Tests...`);
   const portNumber: string | number = process.env.PORT || 3001;
-  return executeSync(`${mochaChromeFilePath} http://localhost:${portNumber} --timeout 60000`, repositoryRootFolderPath).status || 1;
+  return executeSync(`${mochaChromeFilePath} http://localhost:${portNumber} --timeout 60000`, repositoryRootFolderPath).status;
 }
 
 let exitCode = 0;
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -16,14 +16,14 @@ import * as http from "http";
 import * as https from "https";
 import { URLBuilder } from "./url";
 
-const client = axios.create();
+export const axiosInstance = axios.create();
 
 // This hack is still required with 0.19.0 version of axios since axios tries to merge the
 // Content-Type header from it's config["<method name>"] where the method name is lower-case,
 // into the request header. It could be possible that the Content-Type header is not present
 // in the original request and this would create problems while creating the signature for
 // storage data plane sdks.
-client.interceptors.request.use((config: AxiosRequestConfig) => ({
+axiosInstance.interceptors.request.use((config: AxiosRequestConfig) => ({
   ...config,
   method: (config.method as Method) && (config.method as Method).toUpperCase() as Method
 }));
@@ -155,7 +155,7 @@ export class AxiosHttpClient implements HttpClient {
           config.httpAgent = agent.agent;
         }
       }
-      res = await client.request(config);
+      res = await axiosInstance.request(config);
     } catch (err) {
       if (err instanceof axios.Cancel) {
         throw new RestError(err.message, RestError.REQUEST_SEND_ERROR, undefined, httpRequest);
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -16,6 +16,18 @@ import * as http from "http";
 import * as https from "https";
 import { URLBuilder } from "./url";
 
+const client = axios.create();
+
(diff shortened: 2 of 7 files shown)


lib/axiosHttpClient.ts
```diff
@@ -16,6 +16,18 @@ import * as http from "http";
...
+
+// This hack is still required with 0.19.0 version of axios since axios tries to merge the
+// Content-Type header from it's config["<method name>"] where the method name is lower-case,
+// into the request header. It could be possible that the Content-Type header is not present
+// in the original request and this would create problems while creating the signature for
+// storage data plane sdks.
+client.interceptors.request.use((config: AxiosRequestConfig) => ({
+  ...config,
+  method: (config.method as Method) && (config.method as Method).toUpperCase() as Method
+}));
```
amarzavery: This change LGTM.

lib/policies/msRestUserAgentPolicy.browser.ts
```diff
@@ -9,7 +9,7 @@
 import { TelemetryInfo } from "./userAgentPolicy";
 
 interface NavigatorEx extends Navigator {
-  readonly oscpu: string | undefined;
+  readonly oscpu: string;
```
amarzavery: @aby040 - How is this change related to the current problem being fixed?

.scripts/unit.ts
```diff
@@ -75,13 +75,13 @@ function stopTestServer(testServer: ServerProcess): void {
...
   console.log(`Running Node.js Unit Tests...`);
-  return executeSync(`nyc mocha`, repositoryRootFolderPath).status;
+  return executeSync(`nyc mocha`, repositoryRootFolderPath).status || 0;
 }
 
 function runBrowserUnitTests(): number {
   console.log(`Running Browser Unit Tests...`);
   const portNumber: string | number = process.env.PORT || 3001;
-  return executeSync(`${mochaChromeFilePath} http://localhost:${portNumber} --timeout 60000`, repositoryRootFolderPath).status;
+  return executeSync(`${mochaChromeFilePath} http://localhost:${portNumber} --timeout 60000`, repositoryRootFolderPath).status || 0;
```
amarzavery: @aby040 -  Why are you defaulting the exit code to `0`? How is this related to the current problem being fixed in the PR?

lib/policies/msRestUserAgentPolicy.browser.ts
aby040: Sorry, this was accidentally added to this merge request. This change was made to fix some build errors.

.scripts/unit.ts
aby040: Sorry, this was accidentally added to this merge request. This change was made to fix some build errors.

lib/axiosHttpClient.ts
aby040: @amarzavery  Can we have this added to latest 1.x.x release so that the memory leak issue can be resolved in `@azure` packages which are still relying on versions `^1.8.1` of `@azure/ms-rest-js`

daviwil:
Hi @aby040, thanks for this change!  Can you turn on "Allow edits from maintainers" on this PR so that I can make a couple of tweaks before merging?

lib/axiosHttpClient.ts
amarzavery: Yes. If everything goes well, then the next version in the 1.x release train would be 1.8.14. @daviwil is the current owner and should be able to provide more details on this.

princjef:
fwiw it looks like the build failures are due to a newer version of typescript being pulled in by the build than the version used when the original code was written. Probably a side effect of the package-lock.json file being excluded from the repo.

bterlson:
Overall change looks good once we can fix the build issue (probably not an issue in azure-sdk-for-js as we're using a lockfile there locked to node v8 types - a similar thing could be considered for this repo).

lib/axiosHttpClient.ts
aby040: Thanks @amarzavery

aby040:
Sure @daviwil, I have turned on "Allow edits from maintainers". Please let me know if anything else has to be added from my end for this to get merged.

aby040:
@bterlson, sorry I didnt get what you meant by 

> using a lockfile there locked to node v8 types

Shall I make the changes in `.scripts/units.ts` and `lib/policies/msRestUserAgentPolicy.browser.ts` to fix the type issues?

princjef:
Some clarification on next steps would be good. Is the desire to fix the existing build/lockfile issues via a separate issue/PR or to do it as a part of this change?

@bterlson thoughts?

bterlson:
Sorry for the confusing unrelated comment. From my perspective, @aby040's proposal to update .scripts/unit.ts and lib/policies/msRestUserAgentPolicy.browser.ts based on the latest node types seems appropriate, but I'd defer to either @daviwil or @amarzavery

princjef:
@daviwil I've created #377 to try to address the build issues as a standalone change to unblock this PR. I'm also fine if you instead want to do the code changes that were previously a part of this PR. Any thoughts on how you want to proceed so that we can resolve this issue?

daviwil:
I integrated @princjef's build and CI fixes from #377 and fixed the remaining test issues caused by the interceptor move, so this is ready to be merged.  Thanks to @aby040 and @princjef for the help!

daviwil:
`ms-rest-js` 1.8.14 is out: https://www.npmjs.com/package/@azure/ms-rest-js/v/1.8.14

PR #376, opened 2019-10-14
Fix node-fetch require for webpack

### LINKED ISSUE #373: Webpacking ms-rest-js results in  TypeError: fetch is not a function
**Package Version**: 2.0.04
- [x] **nodejs**
   - **nodejs version**: 10.16.0
   - **os name/version**: Windows 10
- [ ] **browser**
    - **name/version**: ____
**Describe the bug**
We are consuming this library via `@azure/storage-blob`. For our scenario, we are running some code on our continuous integration pipeline that uploads files to blob storage. To avoid having to do an NPM install on the CI server, our desire is to webpack our file upload utility and just run the bundled JS file on the build server. However, we are seeing an issue resolving fetch that does not manifest when running the script in a non-bundled context.

**To Reproduce**
1. Create any script that consumes nodeFetchHttpClient.ts. In our case, we are bundling a modified version of the advanced example of the [@azure/storage-blob sdk](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/storage/storage-blob/samples/typescript/advanced.ts) with our blob storage details.
2. Create a webpack bundle with target "node".
3. Run that bundle with `node [bundlefilename].js`

**Expected behavior**
We expect the library to work in a bundled context, without having any issues relating to fetch.

**Screenshots**

**Additional context**
I dug a bit deeper here and I found that folks are having similar issues in other libraries, with the best example being the discussion here: https://github.com/bitinn/node-fetch/issues/450

It appears that in the case for ms-rest-jest, it is this line that is problematic for webpack:
```typescript
if (typeof globalWithFetch.fetch !== "function") {
  const fetch = require("node-fetch");
  globalWithFetch.fetch = fetch;
}
```

It appears to be an issue with commonJS vs ESM module resolution. It seems like webpack would want the import to be require("node-fetch").default. However, I am not sure if this would have implications for folks not consuming this library via webpack.

AnthonyMacKinnon:
fixes #373 

We ran into the same issue as @christiango when using webpack for a server-side node project.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,7 @@
 # Changelog
+## 2.0.5 - 2020-01-07
+- Fix node-fetch bundling when using Webpack (PR [#376](https://github.com/Azure/ms-rest-js/pull/376)).
+
 ## 2.0.4 - 2019-07-30
 - Ensure that a primitive type (string, number, boolean, null, undefined) response body with or without a `bodyMapper` is not flattened.
 
--- a/lib/nodeFetchHttpClient.ts
+++ b/lib/nodeFetchHttpClient.ts
@@ -17,7 +17,7 @@ interface GlobalWithFetch extends NodeJS.Global {
 
 const globalWithFetch = global as GlobalWithFetch;
 if (typeof globalWithFetch.fetch !== "function") {
-  const fetch = require("node-fetch");
+  const fetch = require("node-fetch").default;
   globalWithFetch.fetch = fetch;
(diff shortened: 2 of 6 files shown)


daviwil:
The `autorest.typescript` dependent test failure needs changes to the Azure DevOps scripts (following on from work done in the `1.x` branch).  I'll do that in a separate PR.  Merging this now, thanks @AnthonyMacKinnon!

PR #377, opened 2019-12-06
Lock dependency versions to fix build

princjef:
The 1.x branch of the repository currently does not check in the package-lock.json file. As a result, newer versions of `typescript` and `@types/node` that have been released since the last commit are automatically pulled in and cause the build to fail.

This PR checks the package-lock.json file into source control [per npm recommendations](https://docs.npmjs.com/files/package-lock.json) and locks the versions of `typescript` and `@types/node` to ones that are compatible and allow the build to pass.

### DIFF
--- a/.devops/azure-pipelines-test-dependent-projects.yml
+++ b/.devops/azure-pipelines-test-dependent-projects.yml
@@ -78,7 +78,11 @@ jobs:
       targetPath: $(System.DefaultWorkingDirectory)
   - script: 'mkdir -p $(tempDirectory)'
     displayName: 'mkdir -p $(tempDirectory)'
-  - script: 'git clone https://github.com/daviwil/autorest.typescript.git --single-branch --branch v4-fix-ci autorest.typescript --recursive --depth 1'
+  # Clone back to the last commit where ms-rest-js 1.x was supported (version 4.1.1)
+  - script: |
+      git clone https://github.com/Azure/autorest.typescript.git autorest.typescript --recursive --depth 1
+      cd autorest.typescript
+      git reset --HARD 8b2d805c8df83889aac0fc3875b19d7a5bf13c6f
     workingDirectory: $(tempDirectory)
     displayName: "clone autorest.typescript"
   - script: 'npm install $(Build.SourcesDirectory)/$(msRestAzureJsPackageName)'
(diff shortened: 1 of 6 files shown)


princjef:
It turns out there's also a test which is validating against the exact html returned by https://example.com, which has changed slightly. I've updated the test so that the bodies will match, but there's no guarantee that this test doesn't break in the future.

princjef:
The "Check_Everything" test complains if the version of the package in the code is already published, so i've bumped to 1.8.14 for now to allow it to pass.

The only test that is still failing is the one that tries to use the package in autorest. It's complaining about not being able to find a `primordials` variable. I'm not sure whether the 1.x version of this repo is even expected to work with the latest autorest at this point, so I'm not sure whether there's actually a valid fix to be made for that test.

princjef:
Got past the `primordials` error in the autorest test by downgrading Node.js to 10.x for that specific test ([ref](https://stackoverflow.com/questions/55921442/how-to-fix-referenceerror-primordials-is-not-defined-in-node)), but it's now failing when it runs `npm ci` via lerna because something it's trying to install has no package-lock.json. I'm assuming the issue is related to the autorest repository rather than this one.

daviwil:
Thanks @princjef!  Looking into the Lerna failure now.

daviwil:
Not really sure what's going on with the error that comes back in that step, the `autorest.typecript` repository has a `package_lock.json` in the directory that causes the error, despite what the error message states.  The `autorest.typescript` repo isn't being cloned at a commit or branch and your PR bumps the Node/npm version appropriately so I'm not sure what it's unhappy with.

Can you turn on the "Allow maintainers to edit" so that I can experiment a bit?

princjef:
It should already be enabled. Let me know if you have issues with access

daviwil:
Thanks!  For whatever reason GitHub didn't give me the message that it was turned on :/

daviwil:
Once I get https://github.com/Azure/autorest.typescript/pull/525 merged I'll restart CI on this PR and I believe the problem will be solved.

daviwil:
Finally got everything working again.  This matrix of old versions of `ms-rest-js`, `ms-rest-azure-js`, and `autorest.typescript` is a bit of a nightmare.  Thanks a lot for your help @princjef!

princjef:
No problem, thanks for getting it across the line!

PR #379, opened 2020-01-24
Fix process undefined check to allow for SSR

Livven:
The `!process` check fails if `process` is undefined, e.g. in a browser environment. When doing server-side rendering with something like Gatsby or Next.js (in our case VuePress) the default `proxyPolicy.ts` would be used during the server render, causing an error when running the rendered site in the browser.

### DIFF
--- a/lib/policies/proxyPolicy.ts
+++ b/lib/policies/proxyPolicy.ts
@@ -9,7 +9,7 @@ import { Constants } from "../util/constants";
 import { URLBuilder } from "../url";
 
 function loadEnvironmentProxyValue(): string | undefined {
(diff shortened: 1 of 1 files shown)


xirzec:
The change itself isn't unreasonable, but several files have browser replacements (e.g. base64 encoding) and it's not going to work to try to make the node files all work in the browser.

Is there no way to have SSR honor the browser versions? Or rather, to pack the entire client for the browser and reference that after rendering?

ramya-rao-a:
Closing this PR as we havent heard any answers from the author to the queries posted above in a while

PR #381, opened 2020-04-14
Expand usage of WebResource to an interface

xirzec:
To align with @azure/core-http (see https://github.com/Azure/azure-sdk-for-js/pull/7873) this change widens public interfaces referencing WebResource and HttpHeaders to use interfaces instead. This should prevent future compatibility breaks.

As part of this change I had to do some minor cleanup to the rollup config to avoid a Windows build break.

PR #384, opened 2020-04-23
Use Node 10 when building autorest.typescript to avoid Gulp issue

daviwil:
This change forces the `autorest.typescript` dependent project build to use Node 10 because there's an issue with Gulp when running on Node 12 that causes this error to be thrown after running `npm install`:

```
> @microsoft.azure/autorest.typescript@5.0.2 prepare /home/vsts/work/1/.tmp/autorest.typescript
> gulp install_common && gulp build

fs.js:35
} = primordials;
    ^

[ReferenceError: primordials is not defined]
```

### DIFF
--- a/.devops/azure-pipelines-test-dependent-projects.yml
+++ b/.devops/azure-pipelines-test-dependent-projects.yml
@@ -78,7 +78,7 @@ jobs:
       targetPath: $(System.DefaultWorkingDirectory)
   - script: 'mkdir -p $(tempDirectory)'
     displayName: 'mkdir -p $(tempDirectory)'
-  - script: 'git clone https://github.com/Azure/autorest.typescript.git autorest.typescript --recursive --depth 1'
+  - script: 'git clone --single-branch -b v4x https://github.com/Azure/autorest.typescript.git autorest.typescript --recursive --depth 1'
     workingDirectory: $(tempDirectory)
     displayName: "clone autorest.typescript"
   - script: 'npm install $(Build.SourcesDirectory)/$(msRestJsPackageName)'
--- a/.devops/azure-pipelines-test-dependent-projects.yml
+++ b/.devops/azure-pipelines-test-dependent-projects.yml
@@ -81,12 +81,12 @@ jobs:
   - script: 'git clone --single-branch -b v4x https://github.com/Azure/autorest.typescript.git autorest.typescript --recursive --depth 1'
     workingDirectory: $(tempDirectory)
     displayName: "clone autorest.typescript"
-  - script: 'npm install $(Build.SourcesDirectory)/$(msRestJsPackageName)'
+  - script: 'npm install $(System.DefaultWorkingDirectory)/$(msRestJsPackageName)'
     workingDirectory: $(repoDir)
     displayName: 'npm install @azure/ms-rest-js'
-  - script: 'npm install $(Build.SourcesDirectory)/$(msRestAzureJsPackageName)'
+  - script: 'npm install $(System.DefaultWorkingDirectory)/$(msRestAzureJsPackageName)'
     workingDirectory: $(repoDir)
-    displayName: 'npm install @azure/ms-rest-azure-js' 
+    displayName: 'npm install @azure/ms-rest-azure-js'
   - script: 'cat package.json'
     workingDirectory: $(repoDir)
     displayName: "debug"
--- a/.devops/azure-pipelines-test-dependent-projects.yml
+++ b/.devops/azure-pipelines-test-dependent-projects.yml
@@ -81,12 +81,12 @@ jobs:
   - script: 'git clone https://github.com/Azure/autorest.typescript.git autorest.typescript --recursive --depth 1'
     workingDirectory: $(tempDirectory)
     displayName: "clone autorest.typescript"
-  - script: 'npm install $(Build.SourcesDirectory)/$(msRestAzureJsPackageName)'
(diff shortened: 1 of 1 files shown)


KarishmaGhiya:
@ramya-rao-a so i just pushed a commit to change the order of npm install packages. 
But @daviwil looks like the gulp test task is failing due to some weird reason :(

ramya-rao-a:
Latest errors:

Screen Shot 2020-04-23 at 10 08 44 PM

@daviwil Could this be due to #381? 

Also, @KarishmaGhiya, how did we merge #381 when there was still problems with the pipeline?

ramya-rao-a:
cc @joheredi In case you have any ideas as to why the autorest.typescript test step here is failing

KarishmaGhiya:
@ramya-rao-a @daviwil   so when i checked this pipeline - https://dev.azure.com/azure-public/azsdk/_build/results?buildId=8668&view=results against the PR 381 - it passed. There was an issue with devops and github that the status was not being reported so i manually checked. Was there some other pipeline too that we should have checked before merging that PR?

KarishmaGhiya:
@ramya-rao-a  @daviwil  I am triggering CI on master against the latest commit and this commit -  https://github.com/Azure/ms-rest-js/commit/90053fa355631cfb5e525696e82ec4cd9e3496b9 to understand from when this pipeline has been failing -

Here are the builds - (latest master) - https://dev.azure.com/azure-public/azsdk/_build/results?buildId=8796&view=results
- this commit - https://github.com/Azure/ms-rest-js/commit/90053fa355631cfb5e525696e82ec4cd9e3496b9 - https://dev.azure.com/azure-public/azsdk/_build/results?buildId=8797&view=results

.devops/azure-pipelines-test-dependent-projects.yml
```diff
@@ -76,12 +81,12 @@ jobs:
   - script: 'git clone https://github.com/Azure/autorest.typescript.git autorest.typescript --recursive --depth 1'
```
ramya-rao-a, line 81: @daviwil, @joheredi Could it be that we need to use the v4 branch of the autorest.typescript instead of master?
Because master branch is using core-http and we need the version of the code generator that is using ms-rest-js?

ramya-rao-a: @joheredi Can you try updating the script to use the v4 branch?

ramya-rao-a:
@KarishmaGhiya The TS compilation issue should be solved with the change from @joheredi 
But we seem to be back with the npm install issue. Can you take a look?

.devops/azure-pipelines-test-dependent-projects.yml
daviwil: Good catch on using the `v4` branch, I should have pulled that over already (it's fixed in the `1.x` branch already).

daviwil:
This `npm install` issue is a little too nondeterministic for my tastes...

joheredi:
So it looks like installing `@azure/ms-rest-azure-js` is failing to find `@azure/ms-rest-js`  [(link)](https://dev.azure.com/azure-public/azsdk/_build/results?buildId=8813&view=logs&j=499e71a5-ab92-5cb7-8e77-9c19c2280f63&t=ccb50be2-aeb2-591d-6193-a91449030bf1&l=18)

However one step before it succeeded to install [(link)](https://dev.azure.com/azure-public/azsdk/_build/results?buildId=8813&view=logs&j=499e71a5-ab92-5cb7-8e77-9c19c2280f63&t=e8cc4c5e-2383-5f19-2436-b0d642821504&l=46)

image

Shouldn't it be trying to get it from node_modules instead of `/home/vsts/work/s/azure-ms-rest-js.tgz`?

daviwil:
I think it's trying to install the artifact from the wrong path, just pushed a commit to test my theory.

daviwil:
Ahhh, good catch too!  That's true, the downloaded ms-rest-js artifact needs to be installed first.  I'll update that.

daviwil:
Well, I was able to fix the `npm install` issue but we hit another issue with the `autorest.typescript` tests :/ 

The `npm install` issue was due to the earlier phase when `@azure/ms-rest-azure-js` is built, the automation `npm install`s the tgz for `ms-rest-js` and then its `package.json` gets updated to contain the `tgz` path for that dependency.  The `ms-rest-azure-js` package then gets packaged up, uploaded as an artifact, and then downloaded in the `autorest.typescript` phase.  Because the `package.json` still refers to the `tgz` path from the earlier phase, installation fails.  Fun!

ramya-rao-a:
Alright! And now we are getting the tests run from autorest.typescript, but there is 1 test failing

Screen Shot 2020-04-24 at 10 19 26 PM

@joheredi Any idea about the above test? Keep in mind to check the v4 branch

joheredi:
> Alright! And now we are getting the tests run from autorest.typescript, but there is 1 test failing
> 
> Screen Shot 2020-04-24 at 10 19 26 PM
> 
> @joheredi Any idea about the above test? Keep in mind to check the v4 branch

Looks like it was a transient issue, the pipeline is passing after re-running. 

There is only one error left which is `check-everything` complaining about the package version not being bumped.  I think @daviwil didn't really want to bump, so we may be ready for merge

ramya-rao-a:
Yes, we dont want to update version for this PR. I'll go ahead with the merge.

Thanks for all the work here @daviwil, @KarishmaGhiya and @joheredi

PR #385, opened 2020-04-23
Make sure parsed body is an array when expecting Sequence

joheredi:
Issue:

When an operation spec defines that a response should be of type Sequence (array) but the service actually returns null/undefined instead of an empty list, we end up crashing with the following error as the empty response body gets deserialized as {}
```
TypeError: (_response.parsedBody || []).slice is not a function at flattenResponse (/mnt/c/Users/jepri/scratch/node_modules/@azure/ms-rest-js/dist/msRest.node.js:3518:62)
at /mnt/c/Users/jepri/scratch/node_modules/@azure/ms-rest-js/dist/msRest.node.js:3353:47
at at process._tickCallback (internal/process/next_tick.js:189:7)
```

See Azure/azure-sdk-for-js/issues/8445

**Note**:  `[...(_response.parsedBody || [])]` gets compiled into `(_response.parsedBody || []).slice()`.

Fix:
When we are expecting an array, make sure we have one before proceeding

### DIFF
--- a/lib/serviceClient.ts
+++ b/lib/serviceClient.ts
@@ -556,7 +556,13 @@ export function flattenResponse(_response: HttpOperationResponse, responseSpec: 
     const modelProperties = typeName === "Composite" && (bodyMapper as CompositeMapper).type.modelProperties || {};
     const isPageableResponse = Object.keys(modelProperties).some(k => modelProperties[k].serializedName === "");
     if (typeName === "Sequence" || isPageableResponse) {
-      const arrayResponse = [...(_response.parsedBody || [])] as RestResponse & any[];
+      // We're expecting a sequece(array) make sure that the response body is in the
+      // correct format, if not make it an empty array []
+      const parsedBody =
+        _response.parsedBody && Array.isArray(_response.parsedBody)
+          ? _response.parsedBody
+          : [];
+      const arrayResponse = [...parsedBody] as RestResponse & any[];
 
       for (const key of Object.keys(modelProperties)) {
         if (modelProperties[key].serializedName) {
--- a/lib/serviceClient.ts
+++ b/lib/serviceClient.ts
@@ -559,7 +559,7 @@ export function flattenResponse(_response: HttpOperationResponse, responseSpec: 
       // We're expecting a sequece(array) make sure that the response body is in the
       // correct format, if not make it an empty array []
       const parsedBody =
-        _response.parsedBody && Array.isArray(_response.parsedBody)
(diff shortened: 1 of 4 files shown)


lib/serviceClient.ts
```diff
@@ -551,7 +551,13 @@ export function flattenResponse(_response: HttpOperationResponse, responseSpec:
     const modelProperties = typeName === "Composite" && (bodyMapper as CompositeMapper).type.modelProperties || {};
     const isPageableResponse = Object.keys(modelProperties).some(k => modelProperties[k].serializedName === "");
     if (typeName === "Sequence" || isPageableResponse) {
-      const arrayResponse = [...(_response.parsedBody || [])] as RestResponse & any[];
+      // We're expecting a sequece(array) make sure that the response body is in the
+      // correct format, if not make it an empty array []
+      const parsedBody =
+        _response.parsedBody && Array.isArray(_response.parsedBody)
```
ramya-rao-a, line 557: Array.isArray() on undefined or null returns false, so we can skip the check on parsedBody being falsy here

```suggestion
        Array.isArray(_response.parsedBody)
```

lib/serviceClient.ts
```diff
@@ -551,7 +551,13 @@ export function flattenResponse(_response: HttpOperationResponse, responseSpec:
...
     const isPageableResponse = Object.keys(modelProperties).some(k => modelProperties[k].serializedName === "");
     if (typeName === "Sequence" || isPageableResponse) {
-      const arrayResponse = [...(_response.parsedBody || [])] as RestResponse & any[];
+      // We're expecting a sequece(array) make sure that the response body is in the
+      // correct format, if not make it an empty array []
+      const parsedBody =
+        Array.isArray(_response.parsedBody)
+          ? _response.parsedBody
+          : [];
+      const arrayResponse = [...parsedBody] as RestResponse & any[];
```
ramya-rao-a, line 560: Are there any tests that we can add/update?

joheredi: Added a test for this scenario

ramya-rao-a:
@sasathy1 Can you share which package are you using where you are seeing the issues that would be solved with this PR?

sasathy1:
> @sasathy1 Can you share which package are you using where you are seeing the issues that would be solved with this PR?

We are using @azure/arm-eventhub and @azure/arm-servicebus(both are of version 3.2.0).

ramya-rao-a:
Looks like we have a new error?

Screen Shot 2020-04-27 at 5 17 27 PM

PR #390, opened 2020-04-29
Code change to check if delimiter is tabs-spaces & value is null/undefined

sarangan12:
The reason for this PR has been explained in detail in the gist https://gist.github.com/sarangan12/ddcec9a5789984dba777713855634166. 

@joheredi Please review and approve

@ramya-rao-a FYI.....

PR #391, opened 2020-04-30
Widen RequestPolicyOptions to interface

xirzec:
Work around a compat error with ms-rest-azure-js using an older version of ms-rest-js by accepting an interface version of RequestPolicyOptions that has no private members.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,10 +1,8 @@
 # Changelog
 
-## 2.0.8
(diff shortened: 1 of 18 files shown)


Changelog.md
```diff
@@ -1,4 +1,8 @@
 # Changelog
+
+## 2.0.8
```
daviwil, line 3: So it looks like we haven't even shipped 2.0.7 yet (the pipeline needs to be run manually) so this PR doesn't need to have its version bumped.

xirzec:
One CI failure is going to be addressed by #390

PR #407, opened 2021-01-05
build(deps): updates axios

juliuscc:
Axios has recently fixed a vulnerability. This upgrade makes sure that the safe version of axios is used.

### DIFF
--- a/.devops/azure-pipelines-pr.yml
+++ b/.devops/azure-pipelines-pr.yml
@@ -1,6 +1,3 @@
-trigger:
-- 1.x
-
 jobs:
 - job: Build
   pool:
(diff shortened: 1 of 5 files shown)


joheredi:
@juliuscc thanks a lot for contributing! The changes look good and the tests passed, would you mind bumping the version so that the step `Check_Everything` passes? You just need to run `npm version patch` which will bump the version and commit, and then you'll need to push the change.

ramya-rao-a:
@joheredi I just pushed a commit for the version bump

joheredi:
> @joheredi I just pushed a commit for the version bump

Thanks! I'll run a couple of sanity tests and will merge

joheredi:
Sanity testing went okay, merging after PR build completes

juliuscc:
Thanks!!! ✨

PR #408, opened 2021-01-21
Allow = character in parameter value

### LINKED ISSUE #406: URLQuery.parse fails parsing when = sign in parameter value
Hi!

I believe I have found a problem with parsing query part of URL In latest release available on github: https://github.com/Azure/ms-rest-js/blob/abc4f8f58ef1ac999315877eaf740c2bd0eddcbe/lib/url.ts#L71

Short context: 
I am trying to receive second page of my scale set virtual machines using following command:
```
computeClient.virtualMachineScaleSetVMs.listNext(`https://management.azure.com:443/subscriptions/1111111-1111-1111-1111-1111111111/resourceGroups/ae-workers/providers/Microsoft.Compute/virtualMachineScaleSets/ae-workers-scale-1000/virtualMachines?api-version=2020-06-01&$skiptoken=ODIzZnNkdi01YXNjLTEzc2EtMTI1NS1jYTNhc2QxMmRhMyEvU3Vic2NyaXB0aW9ucy8xMTExMTEtMTExMS0xMTExLTExMTEtMTExMTExMTExL1Jlc291cmNlR3JvdXBzL0FFLVdPUktFUlMvVk1TY2FsZVNldHMvQUUtV09SS0VSUy1TQ0FMRS0xMDAwL1ZNcy8xNDMz==`);
```

Parameter is the `nextLink`, which I received from previous execution of list method.

As you can see returned skiptoken is base64 encoded data with '==' signs at the end. When executing `computeClient.virtualMachineScaleSetVMs.listNext` URLQuery.parse is executed as well.

We can find there something:
```
case "ParameterValue":
                        switch (currentCharacter) {
                            case "=":
                                parameterName = "";
                                parameterValue = "";
                                currentState = "Invalid";
                                break;
```

I believe that this part should be removed - query param value can contain '=' sign. Please tell me if I'm wrong. 
My `$skiptoken` is not correctly parsed and in result I cannot get second page for my scale set virtual machines.

Best,
Krzysztof Kamiński

joheredi:
Currently, ms-rest-js marks a query parameter as invalid and drops it if it contains an `=` character.

According to RFC 3896 "reserved" characters should be encoded ([link](https://tools.ietf.org/html/rfc3986#section-2.2)). However, nodejs and browser's native URL seems to handle = in the parameter value just fine.

With this PR ms-rest-js parsing will allow `=` in a query parameter value.

Fixes #406

### DIFF
--- a/.devops/azure-pipelines-pr.yml
+++ b/.devops/azure-pipelines-pr.yml
@@ -1,3 +1,6 @@
+trigger:
+- master
+
 jobs:
 - job: Build
   pool:
--- a/.devops/azure-pipelines-pr.yml
+++ b/.devops/azure-pipelines-pr.yml
@@ -1,6 +1,3 @@
-trigger:
-- master
-
 jobs:
 - job: Build
   pool:
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,8 @@
 # Changelog
+
+## 2.2.0 - 2021-01-22
+- Allow = character in parameter value (PR [#408](https://github.com/Azure/ms-rest-js/pull/408))
+
 ## 2.1.0 - 2020-10-08
 - Add support for custom http/https agent (PR [#403](https://github.com/Azure/ms-rest-js/pull/403))
 - Fix WebResource clone to include extra settings (Issue [#405](https://github.com/Azure/ms-rest-js/issue/403))
--- a/lib/url.ts
+++ b/lib/url.ts
@@ -102,12 +102,6 @@ export class URLQuery {
 
         case "ParameterValue":
           switch (currentCharacter) {
-          case "=":
-            parameterName = "";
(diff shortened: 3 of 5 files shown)


xirzec:
I'm ok with this change, since it maps to the platform URL parsing behavior

daviwil (approved):
If it works fine in Node and the browser, I don't see any reason not to use it!

ramya-rao-a:
Do we have the same problem in `@azure/core-http`?

joheredi:
> Do we have the same problem in `@azure/core-http`?

I just checked and this is not an issue in core-http

joheredi:
Merging, the only non-passing test is the check-everything. I'm not updating the package version in favor of #410 which is updating the version and we want both fixes to be part of the same release

PR #409, opened 2021-01-22
Fix a build error involving xml2js

deyaaeldeen:
Currently building master fails with the following error:
```lib/util/xml.ts:8:5 - error TS2345: Argument of type '{ explicitArray: boolean; explicitCharkey: boolean; rootName: string | undefined; renderOpts: { pretty: false; }; }' is not assignable to parameter of type 'BuilderOptions'.
  Object literal may only specify known properties, and 'explicitArray' does not exist in type 'BuilderOptions'.

8     explicitArray: false,
      ~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```
because of new changes in `@types/xml2js`. This PR fixes the issue by removing unnecessary arguments from the builder options that were not used anywhere in the constructor. Tests succeeds locally.

### DIFF
--- a/lib/util/xml.ts
+++ b/lib/util/xml.ts
@@ -5,8 +5,6 @@ import * as xml2js from "xml2js";
 
 export function stringifyXML(obj: any, opts?: { rootName?: string }) {
   const builder = new xml2js.Builder({
-    explicitArray: false,
-    explicitCharkey: false,
     rootName: (opts || {}).rootName,
     renderOpts: {
       pretty: false


joheredi (approved):
LGTM

lib/util/xml.ts
```diff
@@ -5,8 +5,6 @@ import * as xml2js from "xml2js";
 
 export function stringifyXML(obj: any, opts?: { rootName?: string }) {
   const builder = new xml2js.Builder({
-    explicitArray: false,
-    explicitCharkey: false,
```
ramya-rao-a, line 9: In `@azure/core-http` which is basically a copy of this project + all changes we wanted to make in Track 2, I see these two properties in use. Are we sure that the Builder does not need these?

ramya-rao-a: I see both of them in https://github.com/Leonidas-from-XIV/node-xml2js#options

deyaaeldeen: they are part of the parser options not the builder options and these are set at lines 18 and 19 here.

ramya-rao-a: Ah ok

ramya-rao-a (approved):

daviwil (approved):

PR #410, opened 2021-01-22
Add TokenCredential adapter

deyaaeldeen:
Enables using identity's TokenCredential using an adapter that converts to `ServiceClientCredentials`.

TODO
- make a release.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,7 @@
 # Changelog
+## 2.1.1 - 2021-01-22
+- Add support for @azure/core-auth's TokenCredential (PR [#410](https://github.com/Azure/ms-rest-js/pull/410))
+
 ## 2.1.0 - 2020-10-08
 - Add support for custom http/https agent (PR [#403](https://github.com/Azure/ms-rest-js/pull/403))
 - Fix WebResource clone to include extra settings (Issue [#405](https://github.com/Azure/ms-rest-js/issue/403))
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,5 @@
 # Changelog
-## 2.1.1 - 2021-01-22
+## 2.2.0 - 2021-01-22
 - Add support for @azure/core-auth's TokenCredential (PR [#410](https://github.com/Azure/ms-rest-js/pull/410))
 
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -50,6 +50,8 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@azure/core-auth": "1.1.4",
```
ramya-rao-a, line 53: Why are we pinning this dependency?

deyaaeldeen: Fixed.

package.json
```diff
@@ -50,6 +50,8 @@
...
+    "@azure/ms-rest-nodeauth": "^3.0.6",
```
ramya-rao-a, line 54: Looks like we need this dependency only for the type `TokenResponse`. Can we make a copy of this instead? Otherwise this heavy dependency will get pulled in unnecessarily in user's code.

daviwil: Agreed

daviwil (approved):

package.json
deyaaeldeen: Fixed.

joheredi (approved):
Looks good! You'll need to bump the package version for the CI build to pass and allow us to release

ramya-rao-a:
> Looks good! You'll need to bump the package version for the CI build to pass and allow us to release

@joheredi Don't we want to include #408 in the release as well? Or do we want to do multiple releases?

Changelog.md
```diff
@@ -1,4 +1,7 @@
 # Changelog
+## 2.1.1 - 2021-01-22
```
ramya-rao-a, line 2: This would be a new feature and not a bug fix. Consider doing a minor version update instead of a patch version update

ramya-rao-a (approved):

Changelog.md
deyaaeldeen: Updated it to 2.2.0.

joheredi:
> > Looks good! You'll need to bump the package version for the CI build to pass and allow us to release
> 
> 
> 
> @joheredi Don't we want to include #408 in the release as well? Or do we want to do multiple releases?

We should release both at the same time. I'll merge the PR and kick off the release once both PRs are merged

Issue #399, opened 2020-08-07
node-sample.js is not a file in dir sample


Vijendra07Kulhade:
**Package Version**: _______
- [x] **nodejs**
   - **nodejs version**:  12.0.0
   - **os name/version**: macOS Cataline version 10.15.6

**Describe the bug**
Tried running `node samples/node-sample.js` as mentioned in the readme. But it seems file itself is not available.
**To Reproduce**
1. `git clone https://github.com/Azure/ms-rest-js.git`
2. Set Subscription ID and Access Token
3. Run `node samples/node-sample.js`
**Expected behavior**
It should login to the account and print the response body.
**Screenshots**

**Additional context**

xirzec:
Could try `npx ts-node samples/node-sample.ts`

jeremymeng:
Fixed by #411. Closing.

PR #411, opened 2021-01-27
Update sample instructions

jeremymeng:
Some changes are ported from core-http.

Addresses #399

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,10 @@ Runtime for isomorphic javascript libraries (that work in the browser and node.j
 
 ### node.js
 - Set the subscriptionId and token
-- Run `node samples/node-sample.js`
+- Run `npx ts-node samples/node-sample.js`
 
 ### In the browser
+- Run `npm run build`
 - Set the subscriptionId and token and then run
 - Open index.html file in the browser. It should show the response from GET request on the storage account. From Chrome type Ctrl + Shift + I and you can see the logs in console.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -5,20 +5,21 @@
...
 - After cloning the repo, execute `npm install`
 
 ## Execution
 
-### node.js
-- Set the subscriptionId and token
-- Run `node samples/node-sample.js`
+### Node.js
+- Set the subscriptionId and token as instructed in `samples/node-samples.ts`
+- Run `npx ts-node samples/node-sample.js`
```
ramya-rao-a, line 18: npx? I don't recall using this elsewhere in our projects. Any particular wins this gives us?

xirzec: `npx` is just an `npm` command to execute a dependency script without installing it globally. e.g. if you install tsc as a devdep in your project you can use `npx tsc` without having to `npm install -g tsc`

jeremymeng: yeah it's convenient for simple commands without needing to add an entry to the package.json script section.  It even installs the package if it's not available locally.

jeremymeng:
The CI failure indicates that we need to bump the version even for non-library changes?

ramya-rao-a:
Since a PR merge is meant to auto publish the package, the CI fails if the version is not incremented. I can override and merge. Can you get an approval from @xirzec ?

ramya-rao-a (approved):

PR #413, opened 2021-01-28
Fix issue of SystemErrorRetryPolicy not retrying

jeremymeng:
by porting fix from https://github.com/Azure/azure-sdk-for-js/pull/8803


lib/policies/systemErrorRetryPolicy.ts
```diff
@@ -104,30 +105,38 @@ function updateRetryData(policy: SystemErrorRetryPolicy, retryData?: RetryData,
...
     // If previous operation ended with an error and the policy allows a retry, do that
-    return utils.delay(retryData.retryInterval)
-        .then(() => policy._nextPolicy.sendRequest(request.clone()))
-        .then(res => retry(policy, request, res, retryData, err))
-        .catch(err => retry(policy, request, operationResponse, retryData, err));
+    try {
+      await utils.delay(retryData.retryInterval);
+      return policy._nextPolicy.sendRequest(request.clone());
+    }
+    catch (err) {
```
xirzec, line 132: can we rename this exception error variable? it shadows the `err` in the function params and makes this more confusing than it needs to be

xirzec (approved):
Looks good with one nit

PR #416, opened 2021-02-01
Move @types/tunnel to devDependencies, not required for building

terry-wen:
@types/tunnel is not required for building, creates unnecessary dependencies when consumed


package.json
```diff
@@ -52,7 +52,6 @@
   "dependencies": {
     "@azure/core-auth": "^1.1.4",
     "@types/node-fetch": "^2.3.7",
```
joheredi, line 54: Would you mind moving @types/node-fetch to dev dependencies as well, please?

joheredi:
Thanks for sending PR. I see that one of the CI steps is failing, it is because it requires a version bump. You can run `npm version patch` and it will bump the patch version.

Would you mind updating the changelog as well for the new 2.2.1: https://github.com/Azure/ms-rest-js/blob/master/Changelog.md you can set `UNRELEASED` instead of the date and I'll update once we release

jeremymeng:
please also update the version in https://github.com/Azure/ms-rest-js/blob/43a03bc7581dcf811447fd96ef9bc091dba0fdf0/lib/util/constants.ts#L10

terry-wen:
Version bumped.

jeremymeng:
We actually had to revert a similar change in core-http because it broke consumer's typescript code https://github.com/Azure/azure-sdk-for-js/pull/9276. 

https://github.com/Azure/azure-sdk-for-js/issues/9278 is logged to track a better fix.

jeremymeng:
We probably should keep @types/node-fetch in `dependencies` @joheredi?

terry-wen:
> We actually had to revert a similar change in core-http because it broke consumer's typescript code [Azure/azure-sdk-for-js#9276](https://github.com/Azure/azure-sdk-for-js/pull/9276).
> 
> [Azure/azure-sdk-for-js#9278](https://github.com/Azure/azure-sdk-for-js/issues/9278) is logged to track a better fix.

While I understand the issue, out of curiosity: if this package on its own doesn't require the dependency for building, how come the consumer couldn't define their own dependency on types/node?

ramya-rao-a:
@jeremymeng I thought that issue was specific to `@types/node`. Didnt realize it applied to types for other packages as well

jeremymeng:
> @jeremymeng I thought that issue was specific to `@types/node`. Didnt realize it applied to types for other packages as well

Yes it's because `@types/node-fetch` has a dependency on `@types/node` so the latter gets installed transitively when `@types/node-fetch` is in the `dependencies` section.  The issue does not affect other packages.

jeremymeng:
> if this package on its own doesn't require the dependency for building

I think `devDependencies` are still required for building this package.

jeremymeng (dismissed):

jeremymeng:
Similar change actually broke ms-rest-azure-js in v1.x #431

ramya-rao-a:
@jeremymeng What is the latest status on this?

jeremymeng:
@ramya-rao-a/@xirzec this is ready to merge. Please review and approve.

xirzec (approved):
Looks reasonable to me

ramya-rao-a (approved):

Issue #394, opened 2020-07-07
ThrottlingRetryPolicy only retries once


joheredi:
See Azure/azure-sdk-for-js/issues/7989 for details

>  FWIW, the API answers a 429 response which the ThrottlingRetryPolicy handles once with a retry, but when that also returns a 429, it doesn't try again. The API says Retry-After 17, but the request is still rejected with a 429 and another Retry-After 17. I wish the API would return a time that is more likely to work out. That is, I would have liked to see the API drive the retries, so that we could easily burst under 100 requests and throttle back based on meaningful responses from the API. Instead, I'm going to have to use p-throttle to proactively throttle and spread 100 requests out over 5 minutes.

jeremymeng:
We could retry more on 429.  Some design questions to consider
- should we provide an option to specify the max retry count
- what value should be the default for max retry count
   - infinite/max integer value, or some small number

jeremymeng:
Fixed by #417. Closing

PR #417, opened 2021-02-02
Keep retrying on 429 responses

jeremymeng:
up to a limit in ThrottlingRetryPolicy.

- Remove `_handleResponse` callback that is used for testing. Use `sinon`
instead to verify that `retry()` has been called.

Addresses #394


lib/policies/throttlingRetryPolicy.ts
```diff
@@ -7,13 +7,23 @@ import { HttpOperationResponse } from "../httpOperationResponse";
 import { Constants } from "../util/constants";
 import { delay } from "../util/utils";
 
-type ResponseHandler = (httpRequest: WebResourceLike, response: HttpOperationResponse) => Promise<HttpOperationResponse>;
 const StatusCodes = Constants.HttpConstants.StatusCodes;
+const DEFAULT_RETRY_COUNT = 10;
```
xirzec, line 11: 10 seems maybe a little high, but I'm not necessarily opposed to it

xirzec (approved):
Changes look good, one minor comment

lib/policies/throttlingRetryPolicy.ts
jeremymeng: I can use 3, which is also used by the other two retry policies.  I was thinking that `Retry-After` is a bit different than other errors in that the service may have a high chance to service the request if the asked delay is respected.

PR #425, opened 2021-02-09
Fix abort signal event handler memory leak

jeremymeng:
The issue is when the same abort signal is passed to multiple requests we keep
adding handlers without releasing them.

Porting fixes from

https://github.com/Azure/azure-sdk-for-js/pull/5649
https://github.com/Azure/azure-sdk-for-js/pull/12038

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Changelog
 
-## 2.2.2 - UNRELEASED
+## 2.2.2 - 2021-02-09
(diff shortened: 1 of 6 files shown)


jeremymeng:
@xirzec @chradek please take a look

chradek (approved):
Looks like a good port! :ship::it:

xirzec (approved):

Issue #428, opened 2021-02-10
Version 1.11.0 missing @azure/core-auth


redjab:
**Package Version**: 1.11.0
- [ ] **nodejs**
   - **nodejs version**: 14
   - **os name/version**: Windows 10
- [ ] **browser**
    - **name/version**: ____

**Describe the bug**

Error: ./node_modules/@azure/ms-rest-js/es/lib/serviceClient.js
Module not found: Error: Can't resolve '@azure/core-auth' in '\node_modules\@azure\ms-rest-js\es\lib'
resolve '@azure/core-auth' in '\node_modules\@azure\ms-rest-js\es\lib'

@azure/core-auth is being listed in this version as a dev dependency instead of a dependency.

jeremymeng:
Should be fixed in v1.11.2 now.  Closing

PR #429, opened 2021-02-10
Move @azure/core-auth to dependencies

jeremymeng:
Addresses #428


deyaaeldeen (approved):
I think we should also have an entry in the changelog.

ramya-rao-a (approved):

jeremymeng:
Looks like @azure/ms-rest-azure-js is broken by the move of @types/tunnels to devDependencies. Looking into it.

jeremymeng:
CI failure due to known issue https://github.com/Azure/autorest.typescript/issues/859. Merging

PR #430, opened 2021-02-10
CI should build ms-rest-azure-js after installing ms-rest-js

jeremymeng:
`npm run test` didn't surface some build errors


daviwil (approved):

jeremymeng:
CI failed because of
1) #431 
2) checking version failed which is fine as we are not changing the library so not incrementing the version number.

PR #433, opened 2021-02-10
[1.x] Revert moving @types/tunnel into devDependencies

jeremymeng:
It is actually required in our public api surface.


jeremymeng:
#434 is likely a better fix.

Issue #431, opened 2021-02-10
ms-rest-js v1.11.0 breaks ms-rest-azure-js


jeremymeng:
```
node_modules/@azure/ms-rest-js/es/lib/axiosHttpClient.d.ts(6,25): error TS7016: Could not find a declaration file for module 'tunnel'. '/home/vsts/work/1/.tmp/ms-rest-azure-js/node_modules/tunnel/index.js' implicitly has an 'any' type.
  Try `npm install @types/tunnel` if it exists or add a new declaration (.d.ts) file containing `declare module 'tunnel';`
```

due to the move of @types/tunnel from dependencies to devDependencies in #422

jeremymeng:
Fixed by #434

PR #434, opened 2021-02-10
Stop depending on tunnel.HttpsOverHttpsOptions for createTunnel()

jeremymeng:
by duplicating the interface. `createTunnel()` is only used internally for testing.

Even though we don't exporse it in public API surface, it is causing problem requiring dependent projects to have dev dependency on @types/tunnel.

### DIFF
--- a/lib/axiosHttpClient.ts
+++ b/lib/axiosHttpClient.ts
@@ -282,7 +282,40 @@ export function createProxyAgent(requestUrl: string, proxySettings: ProxySetting
   return proxyAgent;
 }
 
-export function createTunnel(isRequestHttps: boolean, isProxyHttps: boolean, tunnelOptions: tunnel.HttpsOverHttpsOptions): http.Agent | https.Agent {
+
+// Duplicate tunnel.HttpsOverHttpsOptions to avoid exporting createTunnel() with dependency on @types/tunnel
+// createIunnel() is only imported by tests.
+interface ProxyOptions {
+  host?: string;
+  port?: number;
+  localAddress?: string;
+  proxyAuth?: string;
+  headers: { [key: string]: any };
+}
+
+interface HttpOptions {
+  maxSockets?: number;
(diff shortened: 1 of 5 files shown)


jeremymeng:
The shape of `tunnel.HttpsOverHttpsOptions` has been stable over the years. There's only some minor updates. Hopefully duplicating is fine?

jeremymeng:
> The shape of `tunnel.HttpsOverHttpsOptions` has been stable over the years. There's only some minor updates.

https://github.com/DefinitelyTyped/DefinitelyTyped/commits/master/types/tunnel/index.d.ts

jeremymeng:
"public.Azure.ms-rest-js - test dependent projects" failing because the latest published version has this bug #431

lib/axiosHttpClient.ts
```diff
@@ -282,7 +282,40 @@ export function createProxyAgent(requestUrl: string, proxySettings: ProxySetting
...
+interface HttpOptions {
+  maxSockets?: number;
+  proxy?: ProxyOptions;
+}
+
+interface HttpsOverHttpOptions extends HttpOptions {
+  ca?: Buffer[];
+  key?: Buffer;
+  cert?: Buffer;
+}
```
xirzec, line 305: Can we flatten these two into `HttpsOverHttpsOptions`? We don't use them directly, right?

lib/axiosHttpClient.ts
```diff
@@ -282,7 +282,40 @@ export function createProxyAgent(requestUrl: string, proxySettings: ProxySetting
...
+
+export interface HttpsProxyOptions extends ProxyOptions {
```
xirzec, line 307: same thought here, can we flatten `ProxyOptions` into this interface?

xirzec (approved):
I'm fine with this, though we should try to keep the new surface as flat as possible.

PR #435, opened 2021-02-11
Stop depending on @types/tunnel for createTunnel

jeremymeng:
Porting fix from v1.x branch https://github.com/Azure/ms-rest-js/pull/434


xirzec (approved):
:shipit:

PR #436, opened 2021-02-22
Add Core AsyncIterator Polyfill to ms-rest-js

sarangan12:
This is related to Issue https://github.com/Azure/azure-sdk-for-js/issues/8764.  This PR adds the ```core-asynciterator-polyfill``` dependency to the ```ms-rest-js``` package.

@ramya-rao-a Please review and approve.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 2.2.4 - 2021-02-23
-- Added ```@azure/core-asynciterator-polyfill``` package as a dependency. Please refer [Issue #8764](https://github.com/Azure/azure-sdk-for-js/issues/8764) for further details.
+- Added ```@azure/core-asynciterator-polyfill``` package as a dependency. If the SDKs (that have ```ms-rest-js``` dependency) have async iterables, AND if the SDK user uses a node version before 10.x.x, then the code will error out. This polyfill dependency enables the successful execution of such code.
 
 ## 2.2.3 - 2021-02-10
 - Dependent projects of @azure/ms-rest-js no longer need to have a dev dependency on @types/tunnel.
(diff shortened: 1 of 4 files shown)


Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.2.4 - 2021-02-23
+- Added ```@azure/core-asynciterator-polyfill``` package as a dependency. Please refer [Issue #8764](https://github.com/Azure/azure-sdk-for-js/issues/8764) for further details.
```
xirzec, line 4: I don't feel like the linked issue actually adds much clarity around what importing this polyfill does for consumers

ramya-rao-a: Agreed, lets add a one liner here instead

ramya-rao-a: On second thoughts... do the packages that use ms-rest-js even have async iterators?

sarangan12: Updated Changelog with an one liner. Could you please check if it is acceptable? @ramya-rao-a For the second question, If my memory serves correct, I found this issue only when working with one of the cognitive services which depends on ms-rest-js.

ramya-rao-a: > If my memory serves correct, I found this issue only when working with one of the cognitive services which depends on ms-rest-js.

Let's confirm that. From what I know, the v4 code gen does not use async iterators

sarangan12: On further checking, I could not find any async iterators within cognitive services. It might be related to search sdk with v4 generator whose custom layer had async iterators as [here](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/search/search-documents/src/searchClient.ts#L341). But even then it was dependent only on core-http and not on ms-rest-js. So, the answer is Yes. I believe we do not have async iterators on this.  Any thoughts???

P.S: Now search documents SDK is moved to V6 SDK generators.

ramya-rao-a: I thought so too. So, in that case, we don;t need this change in ms-rest-js

PR #438, opened 2021-03-15
Specify redirect behaviour for indivual requests

paulfaid:
In our usage we want to be able to specify the redirect behaviour for an individual request. I have added the property `redirectLimit` to `RequestPrepareOptions` and `WebResource`. 

It turns out the existing `redirectPolicy` was never executed as the call to node-fetch does not override the default value of redirect='follow'. As it was, node-fetch processed all redirects before the redirectPolicy was executed. I have modified this so that if a redirectLimit is defined, the call to node-fetch will use redirect='manual'

I've also added `url?:string` and `redirected?:boolean` to the response to mimic the behaviour of node fetch. Returning these values seems logical, and in our usage we require them in the response.

I have tried to write some tests that cover the main behavioural changes, but there are no existing tests around the  redirectPolicy. The redirectPolicy code seems to behave differently to other implementations. It would be a good future task to write a set of tests that expect the standard redirect behaviour, maybe based off the node-fetch behavior and documentation, and then to modify the redirectPolicy code to pass these tests.

Happy for all feedback.
Regards
Paul

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,5 @@
 # Changelog
-## 2.5.0 - UNRELEASED
+## 2.5.0 - (2021-05-10)
 
 - Add WebResource.redirectLimit: Limit the number of redirects followed for this request. If set to 0, redirects will not be followed.
 - Port changes to redirect policy from [azure-sdk-for-js](https://github.com/Azure/azure-sdk-for-js/pull/11863/files]
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,6 @@
 # Changelog
+## 2.3.1 - UNRELEASED
+- Add WebResource.RedirectLimit: Limit the number of redirects followed for this request. If set to 0, redirects will not be followed
 
 ## 2.3.0 - 2021-03-29
 - Moving @types dependencies into devdependencies
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 ## 2.3.1 - UNRELEASED
-- Add WebResource.RedirectLimit: Limit the number of redirects followed for this request. If set to 0, redirects will not be followed
+- Add WebResource.redirectLimit: Limit the number of redirects followed for this request. If set to 0, redirects will not be followed
 
 ## 2.3.0 - 2021-03-29
 - Moving @types dependencies into devdependencies
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 ## 2.3.1 - UNRELEASED
-- Add WebResource.redirectLimit: Limit the number of redirects followed for this request. If set to 0, redirects will not be followed
+- Add WebResource.redirectLimit: Limit the number of redirects followed for this request. If set to 0, redirects will not be followed.
 
 ## 2.3.0 - 2021-03-29
 - Moving @types dependencies into devdependencies
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,8 @@
 # Changelog
-## 2.3.1 - UNRELEASED
+## 2.5.0 - UNRELEASED
+
 - Add WebResource.redirectLimit: Limit the number of redirects followed for this request. If set to 0, redirects will not be followed.
+- Port changes to redirect policy from [azure-sdk-for-js](https://github.com/Azure/azure-sdk-for-js/pull/11863/files]
 
 ## 2.4.0 - 2021-04-19
 
--- a/lib/policies/redirectPolicy.ts
+++ b/lib/policies/redirectPolicy.ts
@@ -60,8 +60,19 @@ function handleRedirect(
 
     return policy._nextPolicy
       .sendRequest(request)
-      .then((res) => handleRedirect(policy, res, currentRetries + 1));
+      .then((res) => handleRedirect(policy, res, currentRetries + 1))
+      .then((res) => recordRedirect(res, request.url));
   }
 
   return Promise.resolve(response);
 }
+
+function recordRedirect(response: HttpOperationResponse, redirect: string): HttpOperationResponse {
+  // only record the deepest/last redirect
+  if (!response.redirected) {
+    response.redirected = true;
+    response.url = redirect;
+  }
+  return response;
+}
+
--- a/lib/policies/redirectPolicy.ts
+++ b/lib/policies/redirectPolicy.ts
@@ -44,16 +44,17 @@ function handleRedirect(
   const locationHeader = response.headers.get("location");
   if (
     locationHeader &&
-    (status === 300 || status === 307 || (status === 303 && request.method === "POST")) &&
-    (!policy.maxRetries || currentRetries < policy.maxRetries)
+    (status === 300 || status === 302 || status === 307 || (status === 303 && request.method === "POST")) &&
+    (currentRetries < policy.maxRetries)
   ) {
     const builder = URLBuilder.parse(request.url);
     builder.setPath(locationHeader);
     request.url = builder.toString();
 
-    // POST request with Status code 303 should be converted into a
+    // POST request with Status code 302 and 303 should be converted into a
     // redirected GET request if the redirect url is present in the location header
-    if (status === 303) {
+    // reference: https://tools.ietf.org/html/rfc7231#page-57 && https://fetch.spec.whatwg.org/#http-redirect-fetch
+    if (status === 302 || status === 303) {
       request.method = "GET";
     }
 
--- a/lib/policies/redirectPolicy.ts
+++ b/lib/policies/redirectPolicy.ts
@@ -11,6 +11,27 @@ import {
   RequestPolicyOptionsLike,
 } from "./requestPolicy";
 
+/**
+ * Options for how redirect responses are handled.
+ */
+export interface RedirectOptions {
+  /*
+   * When true, redirect responses are followed.  Defaults to true.
+   */
+  handleRedirects: boolean;
+
+  /*
+   * The maximum number of times the redirect URL will be tried before
+   * failing.  Defaults to 20.
+   */
+  maxRetries?: number;
+}
+
+export const DefaultRedirectOptions: RedirectOptions = {
+  handleRedirects: true,
+  maxRetries: 20,
+};
+
 export function redirectPolicy(maximumRetries = 20): RequestPolicyFactory {
   return {
     create: (nextPolicy: RequestPolicy, options: RequestPolicyOptionsLike) => {
@@ -44,12 +65,14 @@ function handleRedirect(
   const locationHeader = response.headers.get("location");
   if (
     locationHeader &&
-    (status === 300 || status === 302 || status === 307 || (status === 303 && request.method === "POST")) &&
-    (
-      (request.redirectLimit !== undefined && currentRetries < request.redirectLimit)
-      ||
-      (request.redirectLimit === undefined && currentRetries < policy.maxRetries)
-    )) {
+    (status === 300 ||
+      (status === 301 && ["GET", "HEAD"].includes(request.method)) ||
+      (status === 302 && ["GET", "POST", "HEAD"].includes(request.method)) ||
+      (status === 303 && "POST" === request.method) ||
+      status === 307) &&
+    ((request.redirectLimit !== undefined && currentRetries < request.redirectLimit) ||
+      (request.redirectLimit === undefined && currentRetries < policy.maxRetries))
+  ) {
     const builder = URLBuilder.parse(request.url);
     builder.setPath(locationHeader);
     request.url = builder.toString();
@@ -57,8 +80,9 @@ function handleRedirect(
     // POST request with Status code 302 and 303 should be converted into a
     // redirected GET request if the redirect url is present in the location header
     // reference: https://tools.ietf.org/html/rfc7231#page-57 && https://fetch.spec.whatwg.org/#http-redirect-fetch
-    if (status === 302 || status === 303) {
+    if ((status === 302 || status === 303) && request.method === "POST") {
       request.method = "GET";
+      delete request.body;
     }
 
     return policy._nextPolicy
@@ -78,4 +102,3 @@ function recordRedirect(response: HttpOperationResponse, redirect: string): Http
   }
   return response;
 }
-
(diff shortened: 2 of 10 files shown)


jeremymeng:
We fixed a similar issue in https://github.com/Azure/azure-sdk-for-js/pull/11863. I wonder whether if ported here, that fix would work for you @paulfaid?

paulfaid:
The work on the redirectPolicy in azure-sdk-for-js looks great. I think that definetly should be ported here.
However I'd still like to see this PR included, as it adds the ability to overide the redirect behaviour for an individual request. Currently the redirect behaviour is specified when you create the client and applies to all requests. This PR adds an extra capability without changng any existing behaviour. Only a few lines of this would be changed when the rediredctPolicy fix is ported here.

paulfaid:
Actually I might have a go at porting that code.

paulfaid:
Ok, so I've looked at porting that code, and I'm not keen on doing it right now. My reason is that although it's great to see the tests, it doesn't comply very tightly with the  [node-fetch spec](https://fetch.spec.whatwg.org/#http-redirect-fetch). If we switch this code ON for everyone (by including `redirect: "manual"` in `RequestInit` by default), users may start experiencing a change in behaviour and not understand why. In our case we are testing Saml SSO flows, and they include a 302  response to a POST that should be redirected as a GET. Node-fetch handles this whereas this code will not.

Happy to discuss further, but I feel the redirectPolicy should match the node-fetch behaviour very closely before turing it on by default for everyone.

paulfaid:
Is there some issue with this PR that I can address? Happy to discuss and it would be nice to see some progress.

Changelog.md
```diff
@@ -1,4 +1,6 @@
 # Changelog
+## 2.3.1 - UNRELEASED
+- Add WebResource.RedirectLimit: Limit the number of redirects followed for this request. If set to 0, redirects will not be followed
```
xirzec, line 3: nit: it's `redirectLimit`

lib/serviceClient.ts
```diff
@@ -240,12 +240,24 @@ export class ServiceClient {
...
+          // if redirect behaviour is specified on the request, we don't want the client's default redirectPolicy to apply
+          continue;
+        }
+
+        httpPipeline = nextPolicyPipeline;
       }
     }
+
+    if (httpRequest.redirectLimit !== undefined) {
+      httpPipeline = new RedirectPolicy(httpPipeline, this._requestPolicyOptions, httpRequest.redirectLimit);
```
xirzec, line 258: I don't like this implementation. It assumes every request **must** use RedirectPolicy, and it questionably changes the ordering so redirect policy is first.

lib/serviceClient.ts
```diff
@@ -240,12 +240,24 @@ export class ServiceClient {
     let httpPipeline: RequestPolicy = this._httpClient;
     if (this._requestPolicyFactories && this._requestPolicyFactories.length > 0) {
       for (let i = this._requestPolicyFactories.length - 1; i >= 0; --i) {
-        httpPipeline = this._requestPolicyFactories[i].create(
+        const nextPolicyPipeline = this._requestPolicyFactories[i].create(
           httpPipeline,
           this._requestPolicyOptions
         );
+
+        if (httpRequest.redirectLimit !== undefined && nextPolicyPipeline instanceof RedirectPolicy) {
```
xirzec, line 248: I don't understand this if block -- it looks like you're just skipping any redirect policies that are already in the pipeline?

xirzec (changes requested):
I'm not opposed to the functionality desired here, that is to say allowing finer grained control over redirects and the ability to inspect that a request was indeed redirected.

However, I'm definitely opposed to the changes in `serviceClient.ts` and I'm also a bit confused as to how a consumer is supposed to utilize `redirected` and `url` on the response. Is the idea that they would look at `_response` for this information? What would they do with it?

Also kind of wondering which SDKs you are using that you want this support for - `ms-rest-js` is a legacy project to me so I'd like to understand what is causing folks to take new dependencies on it.

lib/serviceClient.ts
paulfaid: The redirectPolicy is **only** applied if a redirectLimit is specificed. If the redirectLimit is specified, then it must be applied as it is the redirectPolicy that counts and limits the redirects.
The ordering is an interesting question. In my testing, placing it first works as expected. If it's exact position in the pipeline becomes important for some othe reason, this code could be extended to support that.

lib/serviceClient.ts
paulfaid: You can only have one redirect policy in the pipeline. If you have more than one, the first one will execute the redirects and any subsequent ones would never see a response with a redirect status code.

paulfaid: Also note: 
>It turns out the existing redirectPolicy was never executed as the call to node-fetch does not override the default value of redirect='follow'. As it was, node-fetch processed all redirects before the redirectPolicy was executed. I have modified this so that if a redirectLimit is defined, the call to node-fetch will use redirect='manual'

paulfaid:
>I'm also a bit confused as to how a consumer is supposed to utilize redirected and url on the response. Is the idea that they would look at _response for this information? What would they do with it?

We are using ms-rest-js in a test project to test OIDC and SAML flows in one of our applications. In some tests we read the redirected url from the response to extract authorization codes and access tokens. 
I'm not sure who originally choose ms-rest-js, but we are using autorest to generate client interfaces between some of our services.

lib/serviceClient.ts
xirzec: It feels strange to enforce this dynamically, before executing each policy. I get that _requestPolicyFactories is opaque, but I don't think the bad design of the policy factories excuses creating further bad design in `ServiceClient`.

In general it feels bad to strongly couple knowledge of a particular pipeline policy into `sendRequest`.

lib/serviceClient.ts
xirzec: Why not just make `redirectPolicy` smart enough to skip itself if `redirectLimit` isn't defined? Why make `ServiceClient` have the burden of knowing the interactions between specific policies?

xirzec:
> 
> We are using ms-rest-js in a test project to test OIDC and SAML flows in one of our applications. In some tests we read the redirected url from the response to extract authorization codes and access tokens.
> I'm not sure who originally choose ms-rest-js, but we are using autorest to generate client interfaces between some of our services.

If you're using AutoRest + ms-rest-js, perhaps we can help you migrate to newer versions of autorest.typescript that use our new (and better supported) core libraries?

lib/serviceClient.ts
paulfaid: That is so obvious now that you've said it. I've removed all of the changes to ServiceClient.

lib/serviceClient.ts
paulfaid: As above, it's obvious now that I should handle this in the redirectPolicy pipeline.

paulfaid:
>If you're using AutoRest + ms-rest-js, perhaps we can help you migrate to newer versions of autorest.typescript that use our new (and better supported) core libraries?

We should investigate migrating to autorest.typescript. It's just tricky to get the time to do this.

lib/policies/redirectPolicy.ts
```diff
@@ -44,23 +44,34 @@ function handleRedirect(
   const locationHeader = response.headers.get("location");
   if (
     locationHeader &&
-    (status === 300 || status === 307 || (status === 303 && request.method === "POST")) &&
-    (!policy.maxRetries || currentRetries < policy.maxRetries)
-  ) {
+    (status === 300 || status === 302 || status === 307 || (status === 303 && request.method === "POST")) &&
+    (request.redirectLimit && currentRetries < request.redirectLimit)) {
```
xirzec, line 48: seems like we no longer consider the policy setting for `maxRetries`? While I'm fine having the request's `redirectLimit` take precedence, I don't think we want to regress support for setting a limit on the policy level.

lib/policies/redirectPolicy.ts
```diff
@@ -44,23 +44,34 @@ function handleRedirect(
...
     const builder = URLBuilder.parse(request.url);
     builder.setPath(locationHeader);
     request.url = builder.toString();
 
-    // POST request with Status code 303 should be converted into a
+    // POST request with Status code 302 and 303 should be converted into a
     // redirected GET request if the redirect url is present in the location header
-    if (status === 303) {
+    // reference: https://tools.ietf.org/html/rfc7231#page-57 && https://fetch.spec.whatwg.org/#http-redirect-fetch
+    if (status === 302 || status === 303) {
```
xirzec, line 56: Reading through the specs confuses me a little; the IETF link says for 302:

```
Note: For historical reasons, a user agent MAY change the request
      method from POST to GET for the subsequent request.
```

Which doesn't seem to motivate strong action. Meanwhile, the fetch spec notes we should switch the method to GET if:

```
actualResponse’s status is 301 or 302 and request’s method is `POST`
actualResponse’s status is 303 and request’s method is not `GET` or `HEAD`
```

But we're not preserving `HEAD` in the 303 case, nor are we handling 301 at all.

xirzec (changes requested):
I'm liking the direction much better now, but I have some implementation quibbles

xirzec:
> 
> We should investigate migrating to autorest.typescript. It's just tricky to get the time to do this.

Fair! Well, let us know in that repository (or in azure-sdk-for-js) if you end up moving in that direction.

lib/policies/redirectPolicy.ts
paulfaid: Yes I found the `for historical reasons` comment not very helpful. I would recommend following the node-fetch implementation as that is what is currently being applied and will cause the minimum disruption to users of this library.

I was initially tempted to port [Azure/azure-sdk-for-js#11863](https://github.com/Azure/azure-sdk-for-js/pull/11863) into this library but choose not to as the behaviour was different from the [node-fetch spec](https://fetch.spec.whatwg.org/#http-redirect-fetch). Specifically regarding the 302 POST redirected as a GET which occurs in one of the SAML flows we are testing.

So where does that leave us with this PR. I would suggest either: 
 1. Respect max retires (even though we know it is not currently used)
 2. Port the `redirectPolicy` changes from `azure-sdk-for-js`, extend the redirect behaviour in line with the `node-fetch` spec and then add the `request.redirectLimit`. 

I'll do option 1 now and see if I have time for option 2.
Any thoughts?

lib/policies/redirectPolicy.ts
paulfaid: I wasn't sure if I should respect maxRetires or remove it:

We're not actually regressing anything, because the call to `node-fetch` always used the deault redirect settings, so node-fetch always followed the redirects before the response got back to this policy. Therefore this policy code **never** actually did anything and `maxReties` never had any effect on the outcome.

The node-fetch redirect behaviour is only overridden if `redirectLimit` is defined, so in other cases the default node-fetch redirect limit of `20` will apply.

xirzec: Sure, I understand that there was a bug before where the policy didn't function as intended, but I don't think it makes sense to continue the buggy behavior in a new way. If we don't want this to be configurable in the policy, I think we should remove it from the signature. This technically is a breaking change, though maybe it doesn't matter as much if it never worked...

lib/policies/redirectPolicy.ts
xirzec: I have a preference for the second option if you're able to, just because keeping this implementation closer to the one in `core-http` will yield easier maintenance in the future.

If the 302 POST becoming a GET is important, I'm fine with adding that behavior, though I think your if check at line 47 should be modified to have

```
(status === 302 && request.method === "POST)"
```

instead of simply `status === 302` since we don't want to convert non-POST 302s to GETs, right?

paulfaid: I agree, option 2 seems the correct thing to do. I'm in no rush to merge this so I'll have a look at implementing option 2 when I get a chance.

paulfaid:
Ok, so I've now ported the redirect policy changes and tests from azure-sdk-for-js.

lib/policies/redirectPolicy.ts
```diff
@@ -11,6 +11,27 @@ import {
...
 } from "./requestPolicy";
 
+/**
+ * Options for how redirect responses are handled.
+ */
+export interface RedirectOptions {
+  /*
+   * When true, redirect responses are followed.  Defaults to true.
+   */
+  handleRedirects: boolean;
```
xirzec, line 21: can we flip this around? I prefer booleans to default to false, so having a "ignoreRedirects" or something would make more sense to me

lib/policies/redirectPolicy.ts
```diff
@@ -44,23 +65,40 @@ function handleRedirect(
...
       .sendRequest(request)
-      .then((res) => handleRedirect(policy, res, currentRetries + 1));
+      .then((res) => handleRedirect(policy, res, currentRetries + 1))
+      .then((res) => recordRedirect(res, request.url));
   }
 
   return Promise.resolve(response);
 }
+
+function recordRedirect(response: HttpOperationResponse, redirect: string): HttpOperationResponse {
```
xirzec, line 97: I spent some time thinking about this, since it's somewhat non-obvious that this methods executes in the context of the penultimate call to `handleRedirect` and thus has the final url, but I can't think of a better way to do it that isn't more code.

Partly I wish we had written this method using async/await and used a while loop to do following instead of recursion, but I won't force that much of a rewrite upon you. 😄

xirzec (approved):
I think these changes look good! It's been a long road, but I appreciate your commitment @paulfaid! 

@jeremymeng Do you have any comments/suggestions about where we've landed?

lib/policies/redirectPolicy.ts
paulfaid: I agree with you, I prefer things that default to false, however I coped this from azure-sdk-for-js, so I left it like this in the interests of compatability. When anyone switches from this library to azure-sdk-for-js, we don't want them to have to invert any redirect options in their calling code. I would suggest we keep this as it is.

lib/policies/redirectPolicy.ts
paulfaid: Add a few words to the comment to help clarify.

paulfaid:
Happy to merge?

jeremymeng:
> @jeremymeng Do you have any comments/suggestions about where we've landed?

This looks great. Thanks a lot @paulfaid!

jeremymeng:
v2.5.0 has been published: https://www.npmjs.com/package/@azure/ms-rest-js/v/2.5.0

PR #441, opened 2021-03-22
Port some proxy enhancements from core-http

jeremymeng:
- NO_PROXY and ALL_PROXY support
- username and password in proxy url string

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 ## 2.3.0 - UNRELEASED
 - Moving @types dependencies into devdependencies
+- Port NO_PROXY and ALL_PROXY support.
+- Port username/password support in proxy url string.
 
 ## 2.2.3 - 2021-02-10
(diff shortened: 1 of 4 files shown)


Changelog.md
```diff
@@ -2,6 +2,8 @@
 
 ## 2.3.0 - UNRELEASED
 - Moving @types dependencies into devdependencies
+- Port NO_PROXY and ALL_PROXY support.
+- Port username/password support in proxy url string.
```
ramya-rao-a, line 6: For ms-rest-js users "porting" wouldnt make sense, they would not have the context. 

```suggestion
- Add NO_PROXY and ALL_PROXY support.
- Add username/password support in proxy url string.
```

xirzec (approved):

Changelog.md
```diff
@@ -2,6 +2,8 @@
 
 ## 2.3.0 - UNRELEASED
```
ramya-rao-a, line 3: When are we planning to release this?

PR #442, opened 2021-03-26
Update WebResource.prepare to also copy streamResponseBody

### LINKED ISSUE #393: HttpOperationResponse.readableStreamBody always null
**Package Version**: 2.0.4
- [x ] **nodejs**
   - **nodejs version**: 12.15
   - **os name/version**: OSX 10.14.6

**Describe the bug**
When using ServiceClient to send a request, the HTTPOperationResponse always has readableStreamBody set to null. I am trying to use the ServiceClient to download a large storage blob, and it always converts the entire body to a string, and won't let me use the readable stream to pipe to a file.

**To Reproduce**
Here's a snippet  example of what I'm trying to do:
```ts
const creds = await AzureCliCredentials.create({
    resource: "https://storage.azure.com"
});
const client = new ServiceClient(creds);
const request: RequestPrepareOptions = {
            url: "https://example.blob.core.windows.net/example/myblob.tar.gz",
            method: "GET",
            headers: {"X-Ms-Version":"2018-11-09"},
};
const res = await client.sendRequest(request)
await new Promise((resolve, reject) => {
            const writeStream = fs.createWriteStream(path)
            .on('error', reject)
            .on('close', resolve)

            res.readableStreamBody!.pipe(writeStream)
            .on('error', reject)
})
```

**Expected behavior**
I would expect res.readableStreamBody to not be null and the large response body to be piped to the outputFile. Instead readableStreamBody is null and nothing is piped

**Additional context**
I may be missing something, but have been reading through all the docs and trying to find examples, but can't figure out why readableStreamBody is null

bwateratmsft:
Fixes #393. If a simple options object is passed to `sendRequest`, [this code](https://github.com/Azure/ms-rest-js/blob/4208098ab146ad6d37611306c01aac3c93db0941/lib/serviceClient.ts#L227-L238) tries to normalize it into a common `WebResource` object via `WebResource.prepare`. This method does not currently copy in `streamResponseBody` from the options, but it should, so this PR fixes that.

### DIFF
--- a/lib/webResource.ts
+++ b/lib/webResource.ts
@@ -471,6 +471,7 @@ export class WebResource {
     this.abortSignal = options.abortSignal;
     this.onDownloadProgress = options.onDownloadProgress;
     this.onUploadProgress = options.onUploadProgress;
+    this.streamResponseBody = options.streamResponseBody;
 
     return this;
   }
--- a/lib/webResource.ts
+++ b/lib/webResource.ts
@@ -606,6 +606,7 @@ export interface RequestPrepareOptions {
   abortSignal?: AbortSignalLike;
   onUploadProgress?: (progress: TransferProgressEvent) => void;
   onDownloadProgress?: (progress: TransferProgressEvent) => void;
+  streamResponseBody?: boolean;
 }
 
 /**


xirzec (approved):
LGTM

jeremymeng (approved):
Thanks for the fix!

ramya-rao-a:
@jeremymeng Can we add a changelog entry for this and then release an update on Monday if not today?

jeremymeng:
> @jeremymeng Can we add a changelog entry for this and then release an update on Monday if not today?

Yes will do it on Monday.

jeremymeng:
@ramya-rao-a opened PR #443

Issue #445, opened 2021-04-17
AzureIdentityCredentialAdapter not available to users


jkpl:
**Package Version**: 2.3.0
- [ ] **nodejs**
   - **nodejs version**: 14.15.4
   - **os name/version**: Ubuntu 20.04 (WSL)
 
**Describe the bug**
`AzureIdentityCredentialAdapter` is not available to users even though it's included in the official documentation page: <https://docs.microsoft.com/en-us/javascript/api/@azure/ms-rest-js/azureidentitycredentialadapter?view=azure-node-latest>

**Additional context**
I'm using Azure SDK libraries that use both the `azure-identity` and the `ms-rest-*` libraries for authentication. Instead of configuring both authentication types, I would like to use just one auth method. Since `azure-identity` seems to the direction where the libraries will move to, I was wondering if I could wrap it and use it with the libraries that depend on `ms-rest-*`. It seems that the `AzureIdentityCredentialAdapter` does just that! Unfortunately, it seems that it's not exposed to the library users but it's only used internally by `ServiceClient` even though its documentation is available online. Could you please export the `AzureIdentityCredentialAdapter` class?

jkpl:
As a workaround, I've copied the `AzureIdentityCredentialAdapter` module to my project for now.

ramya-rao-a:
Thanks for reporting @jkpl!
This is a fair ask and we have a PR out to do this: https://github.com/Azure/ms-rest-js/pull/446

I wanted to take this opportunity to share that we are updating the packages that depend on ms-rest-* to support the credentials from `@azure/identity` package directly. That is why we created this adapter in the first place :)

You can subscribe to https://github.com/Azure/azure-sdk-for-js/issues/14922 to track our efforts on this.

jkpl:
Awesome! Thanks @ramya-rao-a! :tada:

jeremymeng:
@jkpl We just published @azure/ms-rest-js 2.4.0 which contains a fix to this issue. Please give it a try!
Thanks @ramya-rao-a for the PR!

ramya-rao-a:
Thanks @jeremyment for the prompt release!

jkpl:
Works perfectly! Thanks for the quick response!

PR #446, opened 2021-04-19
Expose AzureIdentityCredentialAdapter in the public API

ramya-rao-a:
The `AzureIdentityCredentialAdapter` allows users of this package to make use of credentials from `@azure/identity`.
When it was added, the plan was to then update the auto generated packages that depend on this package to have a constructor overload that would take the `TokenCredential` interface. As it would take a while to update over 150 packages to have this change, it is beneficial to expose this adapter for folks who may want to make use of it sooner.

Case in point: #445

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,60 +1,78 @@
 # Changelog
 
+## 2.4.0 - 2021-04-19
+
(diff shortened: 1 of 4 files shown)


jeremymeng (approved):
LGTM

PR #447, opened 2021-05-04
Speculative fix for DOM manipulation using standalone xmldom library

### LINKED ISSUE #401: `DOMParser is not defined` error when initializing client in Web Worker
**Package Version**: 2.0.8
- [ ] **nodejs**
   - **nodejs version**: ______
   - **os name/version**: _____
- [x] **browser**
    - **name/version**: Chrome Version 85.0.4183.121

**Describe the bug**
When initializing a ms-rest-js client within a web worker, an error is thrown: `DOMParser is not defined`. DOMParser is not available within a web worker. This prevents the client from being used from within a web worker.

**To Reproduce**
1. Create a web worker using code similar to the following (substitute `ClientName`, `token`, and `customBaseUri` for valid values--`ClientName` should be a client derived from `msRest.ServiceClient`; the other values might be optional depending on the client that is tested)

```
onmessage = () => {
  const client = new ClientName(new TokenCredentials(token), {
    baseUri: "customBaseUri"
  });
};
```

2. Send a message to the web worker to invoke this code
3. Note error logged in browser console:

```
Uncaught ReferenceError: DOMParser is not defined
    at Module.../../node_modules/@azure/ms-rest-js/es/lib/util/xml.browser.js (xml.browser.js:3)
    at __webpack_require__ (bootstrap:19)
    at Module.../../node_modules/@azure/ms-rest-js/es/lib/policies/deserializationPolicy.js (deserializationPolicy.js:1)
    at __webpack_require__ (bootstrap:19)
    at Module.../../node_modules/@azure/ms-rest-js/es/lib/serviceClient.js (serviceClient.js:1)
    at __webpack_require__ (bootstrap:19)
    at Module.../../node_modules/@azure/ms-rest-js/es/lib/msRest.js (msRest.js:1)
    at __webpack_require__ (bootstrap:19)
```

**Expected behavior**
The client should initialize successfully within a web worker.

**Screenshots**
Screenshot of error in browser console:
image

**Additional context**
N/A

rathboma:
Replaces browser-built-in DOM libs with `xmldom`. Should fix issues using library in webworkers. Testing now.

Also, you guys really should commit a package-lock.json or a yarn.lock file to make sure your resolved dependencies don't change between machines. I've added a yarn.lock, because I use yarn, but that's up to you.

Fix #401


jeremymeng:
@rathboma Thanks for the PR!  However, in general we try to avoid adding additional dependencies for browser users as it unnecessarily increases the bundle size for users who don't need XML feature. This is why we suggest polyfilling the DOM api as a workaround.

jeremymeng:
@rathboma a polyfill workaround would be similar to the one in https://github.com/Azure/azure-sdk-for-js/blob/589ce99c13a16de07dd8e9dc2117cd3e44666d4d/samples/web-workers/js/polyfill.worker.js.  I am closing this PR.  Please let us know if you have any further questions.

PR #448, opened 2021-05-05
Use self instead of window when referring to the global object

maorleger:
## What

- Use self instead of window when referring to the global object

## Why

- This came up recently in a few cases: https://github.com/Azure/azure-sdk-for-js/issues/11067 and in this discussion: https://github.com/Azure/azure-sdk-for-js/pull/13232#discussion_r557793538
- window is not always the global object, but self's reference changes depending on the context and will point to the right global object

We made this change in azure-sdk-for-js but we should make it here as well.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.4.1 - (unreleased)
+
+- Use `self` instead of `window` in order to support web workers.
+
 ## 2.4.0 - 2021-04-19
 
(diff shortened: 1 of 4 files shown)


jeremymeng (commented):
Looks good to me. Can you please use today's date in the changelog? I can trigger a release after this is merged.

jeremymeng:
The version number needs to be updated
- package.json
- lib/util/constants.ts

maorleger:
@jeremymeng Thanks! First time making changes in this repo :smile: 

Feel free to let me know if there's anything else I need to change (and I'll keep an eye on CI)

jeremymeng (approved):

PR #451, opened 2021-06-06
Copy attributes back post de-serializing array

### LINKED ISSUE #450: Array flattening in deserializer loses previously de-serialized attributes
Referencing the following:
https://github.com/Azure/ms-rest-js/blob/ef37d563d7f604cab8a151c8cc50f064bddfc1b4/lib/serializer.ts#L738

This code blows away any previously deseriaized attributes on the instance. This causes a loss of information such as the odata.* attributes in the following model. 

```
    var PolicyEventsQueryResults = {
        serializedName: "PolicyEventsQueryResults",
        type: {
            name: "Composite",
            className: "PolicyEventsQueryResults",
            modelProperties: {
                odatacontext: {
                    serializedName: "@odata\\.context",
                    type: {
                        name: "String"
                    }
                },
                odatacount: {
                    serializedName: "@odata\\.count",
                    constraints: {
                        InclusiveMinimum: 0
                    },
                    type: {
                        name: "Number"
                    }
                },
                odatanextLink: {
                    serializedName: "@odata\\.nextLink",
                    type: {
                        name: "String"
                    }
                },
                value: {
                    serializedName: "",
                    type: {
                        name: "Sequence",
                        element: {
                            type: {
                                name: "Composite",
                                className: "PolicyEvent",
                                additionalProperties: {
                                    type: {
                                        name: "Object"
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
```

A short-term work-around is to move the value property to the start of the model, but a proper fix would be to either sort the order of properties being parsed to do the array generation first, or to save and re-apply any existing attributes.

mgranell:
Fixes #450

### DIFF
--- a/lib/serializer.ts
+++ b/lib/serializer.ts
@@ -735,7 +735,15 @@ function deserializeCompositeType(
       // paging
       if (Array.isArray(responseBody[key]) && modelProps[key].serializedName === "") {
         propertyInstance = responseBody[key];
-        instance = serializer.deserialize(propertyMapper, propertyInstance, propertyObjectName);
+        const arrayInstance = serializer.deserialize(propertyMapper, propertyInstance, propertyObjectName);
+        // Copy over any properties that have already been added into the instance, where they do
+        // not exist on the newly de-serialized array
+        Object.entries(instance).forEach(([key, value]) => {
+          if (!arrayInstance.hasOwnProperty(key)) {
+            arrayInstance[key] = value;
+          }
+        });
+        instance = arrayInstance
       } else if (propertyInstance !== undefined || propertyMapper.defaultValue !== undefined) {
         serializedValue = serializer.deserialize(
           propertyMapper,
--- a/lib/serializer.ts
+++ b/lib/serializer.ts
@@ -738,11 +738,11 @@ function deserializeCompositeType(
         const arrayInstance = serializer.deserialize(propertyMapper, propertyInstance, propertyObjectName);
         // Copy over any properties that have already been added into the instance, where they do
         // not exist on the newly de-serialized array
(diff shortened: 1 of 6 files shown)


lib/serializer.ts
```diff
@@ -735,7 +735,15 @@ function deserializeCompositeType(
       // paging
       if (Array.isArray(responseBody[key]) && modelProps[key].serializedName === "") {
         propertyInstance = responseBody[key];
-        instance = serializer.deserialize(propertyMapper, propertyInstance, propertyObjectName);
+        const arrayInstance = serializer.deserialize(propertyMapper, propertyInstance, propertyObjectName);
+        // Copy over any properties that have already been added into the instance, where they do
+        // not exist on the newly de-serialized array
+        Object.entries(instance).forEach(([key, value]) => {
```
xirzec, line 741: nit
```suggestion
        for (const [key, value] of Object.entries(instance)) {
```

mgranell:
> nit
```suggestion
        for (const [key, value] of Object.entries(instance)) {
```

lib/serializer.ts
mgranell: Thanks, updated

ramya-rao-a:
@jeremymeng, @xirzec Do we have a similar issue in `@azure/core-http` or the newer Azure Core packages?

jeremymeng:
> @jeremymeng, @xirzec Do we have a similar issue in `@azure/core-http` or the newer Azure Core packages?

I am looking into it.

jeremymeng (approved):

jeremymeng:
@deyaaeldeen @xirzec are you good with the fix?

xirzec (approved):

Issue #455, opened 2021-07-11
Security vulnerability for uuid@3.x.x


sindilevich:
**Is your feature request related to a problem? Please describe.**
Currently, installing `@azure/ms-rest-js@latest`, produces the following message:

> deprecated uuid@3.4.0: Please upgrade to version 7 or higher. Older versions may use Math.random() in certain circumstances, which is known to be problematic. See https://v8.dev/blog/math-random for details.

Please upgrade to the latest version of `uuid` for better security/stronger randomness.

**Describe the solution you'd like**
Upgrade the `uuid` dependency to v7+.

**Describe alternatives you've considered**
No alternatives were considered, since the report is about an outdated and potentially vulnerable dependency: `uuid@3.x.x`.

**Additional context**
> deprecated uuid@3.4.0: Please upgrade to version 7 or higher. Older versions may use Math.random() in certain circumstances, which is known to be problematic. See https://v8.dev/blog/math-random for details.

deyaaeldeen:
Hi @sindilevich, thanks a lot for the report! I updated uuid to v8 in https://github.com/Azure/ms-rest-js/pull/456.

sindilevich:
@deyaaeldeen, thank you so much!

deyaaeldeen:
@sindilevich I am glad to be able to help! I made a release for v2.5.3 that you can find here: https://www.npmjs.com/package/@azure/ms-rest-js/v/2.5.3

PR #456, opened 2021-07-12
Upgrade uuid dependency to v8

deyaaeldeen:
From https://github.com/Azure/ms-rest-js/issues/455:
> Currently, installing @azure/ms-rest-js@latest, produces the following message:

> deprecated uuid@3.4.0: Please upgrade to version 7 or higher. Older versions may use Math.random() in certain circumstances, which is known to be problematic. See https://v8.dev/blog/math-random for details.

So this PR upgrades dep on uuid to latest version. Fixes https://github.com/Azure/ms-rest-js/issues/455

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.5.3 - (2021-07-12)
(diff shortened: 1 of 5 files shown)


jeremymeng (approved):
Looks good. Thanks for fixing it!

PR #457, opened 2021-07-27
Adds property baseUri on the ServiceClientOptions

sadasant:
Added a new property `baseUri` on the `ServiceClientOptions`, so that when a `TokenCredential` is sent to `ServiceClient`, it now automatically sets the scope of future token requests based on the `options.baseUri` property.

This should address https://github.com/Azure/azure-sdk-for-js/issues/15945

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.6.0 - (2021-07-29)
+## 2.6.0 - (2021-08-02)
 
 - Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`. This allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.5.4 - (2021-07-29)
+- When a `TokenCredential` is sent to `ServiceClient`, it now automatically sets the scope of future token requests based on the `baseUri` property.
+
 ## 2.5.3 - (2021-07-12)
 - Updated the dependency on the uuid package to v8 (PR [456](https://github.com/Azure/ms-rest-js/pull/456))
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.5.4 - (2021-07-29)
+## 2.6.0 - (2021-07-29)
 
 - Added a new property `baseUri` on the `ServiceClientOptions`, so that when a `TokenCredential` is sent to `ServiceClient`, it now automatically sets the scope of future token requests based on the `options.baseUri` property.
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,7 +3,7 @@
 ## 2.6.0 - (2021-08-16)
 
 - Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`.
-  - For `baseUri` that happen to be known Azure resource management endpoints, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
+  - For `baseUri` that happen to be known Azure resource manager endpoints, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
 
 ## 2.5.3 - (2021-07-12)
 - Updated the dependency on the uuid package to v8 (PR [456](https://github.com/Azure/ms-rest-js/pull/456))
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,8 @@
 
 ## 2.6.0 - (2021-08-02)
 
-- Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`. This allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
+- Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`.
+  - For `baseUri` that happen to be known Azure management clouds, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
 
 ## 2.5.3 - (2021-07-12)
 - Updated the dependency on the uuid package to v8 (PR [456](https://github.com/Azure/ms-rest-js/pull/456))
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## 2.5.4 - (2021-07-29)
-- When a `TokenCredential` is sent to `ServiceClient`, it now automatically sets the scope of future token requests based on the `baseUri` property.
+
+- Added a new property `baseUri` on the `ServiceClientOptions`, so that when a `TokenCredential` is sent to `ServiceClient`, it now automatically sets the scope of future token requests based on the `options.baseUri` property.
 
 ## 2.5.3 - (2021-07-12)
 - Updated the dependency on the uuid package to v8 (PR [456](https://github.com/Azure/ms-rest-js/pull/456))
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,7 +3,7 @@
 ## 2.6.0 - (2021-08-16)
 
 - Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`.
-  - For `baseUri` that happen to be known Azure management clouds, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
+  - For `baseUri` that happen to be known Azure resource management endpoints, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
 
 ## 2.5.3 - (2021-07-12)
 - Updated the dependency on the uuid package to v8 (PR [456](https://github.com/Azure/ms-rest-js/pull/456))
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.6.0 - (2021-08-02)
+## 2.6.0 - (2021-08-16)
 
 - Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`.
   - For `baseUri` that happen to be known Azure management clouds, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 ## 2.6.0 - (2021-07-29)
 
-- Added a new property `baseUri` on the `ServiceClientOptions`, so that when a `TokenCredential` is sent to `ServiceClient`, it now automatically sets the scope of future token requests based on the `options.baseUri` property.
+- Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`. This allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
 
 ## 2.5.3 - (2021-07-12)
 - Updated the dependency on the uuid package to v8 (PR [456](https://github.com/Azure/ms-rest-js/pull/456))
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.6.0 - (2021-08-16)
+## 2.6.0 - (2021-08-18)
 
 - Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`.
   - For `baseUri` that happen to be known Azure resource manager endpoints, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
--- a/lib/serviceClient.ts
+++ b/lib/serviceClient.ts
@@ -54,7 +54,10 @@ import { agentPolicy } from "./policies/agentPolicy";
 import { proxyPolicy, getDefaultProxySettings } from "./policies/proxyPolicy";
 import { throttlingRetryPolicy } from "./policies/throttlingRetryPolicy";
 import { Agent } from "http";
-import { AzureIdentityCredentialAdapter } from "./credentials/azureIdentityTokenCredentialAdapter";
+import {
+  AzureIdentityCredentialAdapter,
+  azureResourceManagerEndpoints,
+} from "./credentials/azureIdentityTokenCredentialAdapter";
 
 /**
  * HTTP proxy settings (Node.js only)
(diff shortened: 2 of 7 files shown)


lib/serviceClient.ts
```diff
@@ -187,7 +192,10 @@ export class ServiceClient {
 
     let serviceClientCredentials: ServiceClientCredentials | undefined;
     if (isTokenCredential(credentials)) {
-      serviceClientCredentials = new AzureIdentityCredentialAdapter(credentials);
+      serviceClientCredentials = new AzureIdentityCredentialAdapter(
+        credentials,
+        options?.baseUri ? `${options.baseUri}/.default` : undefined
```
chradek, line 197: Should we also set `this.baseUri` to `options.baseUri` if it exists? It's a protected member on this class, but I don't feel too strongly about it.

sadasant: Thanks! I thought about this, then I forgot to do it 🤦‍♂️

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.5.4 - (2021-07-29)
```
ramya-rao-a, line 3: Since we are adding a new feature, we should increment the minor version instead.

lib/serviceClient.ts
```diff
@@ -143,6 +143,11 @@ export interface ServiceClientOptions {
    * HTTP and HTTPS agents which will be used for every HTTP request (Node.js only).
    */
   agentSettings?: AgentSettings;
+  /**
+   * If specified, this is the base URI that requests will be made against for this ServiceClient.
+   * If it is not specified, then all OperationSpecs must contain a baseUrl property.
```
ramya-rao-a, line 148: I see that we copied this from the docs for the corresponding protected property in ServiceClient. Now that we are adding one more use for it, that should be documented here and on that property in ServiceClient as well

ramya-rao-a (changes requested):

ramya-rao-a:
Any concerns with subsequent interfaces extending ServiceClientOptions and redeclaring the `baseUri` property?

lib/serviceClient.ts
sadasant: How does this look? https://github.com/Azure/ms-rest-js/pull/457/commits/557e53794a930b89c76eea396cb2a030724c5d73#diff-9f4919aa6f25f2ebd2d7e5f17c0d2c706cbc31691fc28e197ab9fd662a316763R147-R153

sadasant:
@ramya-rao-a I don’t know of use cases like that. This approach is simpler than others we’ve talked, but it’s less considerate of edge cases. We could have something smarter, like a function that lets you replace the scope on getToken call. cc: @chradek

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.6.0 - (2021-07-29)
+
+- Added a new property `baseUri` on the `ServiceClientOptions`, so that when a `TokenCredential` is sent to `ServiceClient`, it now automatically sets the scope of future token requests based on the `options.baseUri` property.
```
ramya-rao-a, line 5:
```suggestion
- Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`. This allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
```

sadasant:
Oh I understood now that you meant if nothing breaks if the baseUri property is re-declared on constructors (I thought you meant late assignments). Testing to be sure!

chradek:
> Any concerns with subsequent interfaces extending ServiceClientOptions and redeclaring the baseUri property?

In our repo there are several instances where ServiceClientOptions is extended and the sub-interface defines `baseUri?: string`. This is compatible with the change.
- ApplicationInsightsDataClientOptions
- LocalSearchClientOptions
- LogAnalyticsClientOptions
- ServiceFrabricClientOptions

However I did notice that more common that `baseUri` was `endpoint`. `endpoint` is used to help construct the `baseUri` for the services that use it (cognitiveservices-entitysearch is an example).  
https://github.com/Azure/azure-sdk-for-js/blob/ce2c1f5310bf0e71dbf8bc2b29d5731a9dfd961f/sdk/cognitiveservices/cognitiveservices-entitysearch/src/entitySearchClientContext.ts#L40-L48
```ts
    super(credentials, options);

    this.endpoint = 'https://api.cognitive.microsoft.com';
    this.baseUri = "{Endpoint}/bing/v7.0";
    this.requestContentType = "application/json; charset=utf-8";
    this.credentials = credentials;
    if (options.endpoint !== null && options.endpoint !== undefined) {
      this.endpoint = options.endpoint;
    }
```

So @sadasant, I don't know if you'll also need to handle `endpoint`. With regards to the IdentityAdapter, where does the `scope` come from if you don't override it in the constructor?

sadasant:
@chradek In the `AzureIdentityCredentialAdapter` (part of this repo) the scope is either the default one, or is assigned at the constructor.

On Identity, the scope is received on each `getToken` call, which is why the translation is a bit weird. I think that if we wanted to do something robust that could work for every possible use case, we would allow the adapter to overwrite the constructor scope on getToken. However, since our target audience for this adapter is the generated SDKs, I believe the current approach of this PR to be a reasonable one, since `baseUri` is always sent through by the children classes.

Let me know if I’m missing something 🤔

sadasant:
@chradek for the case you presented here: https://github.com/Azure/azure-sdk-for-js/blob/ce2c1f5310bf0e71dbf8bc2b29d5731a9dfd961f/sdk/cognitiveservices/cognitiveservices-entitysearch/src/entitySearchClientContext.ts#L40-L48

I believe that scope is specific to the service version they use. For them, the default scope is likely sufficient, unless they’re working on external clouds.

But to confirm I would need to test it, or speak with somebody owning the cognitive search client. Let me see if I can find someone.

chradek:
> I believe that scope is specific to the service version they use. For them, the default scope is likely sufficient, unless they’re working on external clouds.

For what it's worth, I don't think what cognitive services is doing with `endpoint` precludes us from exposing `baseUri`.
It's possible that for cognitive services, we'd have to construct the baseUri to pass to the ServiceClient if the user specified a custom endpoint, but we'd need to know what the scopes look like for non-public clouds for those services. But that can be done if needed and I don't think has to be part of this PR.

lib/serviceClient.ts
```diff
@@ -151,8 +161,13 @@ export interface ServiceClientOptions {
...
   /**
-   * If specified, this is the base URI that requests will be made against for this ServiceClient.
-   * If it is not specified, then all OperationSpecs must contain a baseUrl property.
+   * If specified:
+   * - This `baseUri` becomes the base URI that requests will be made against for this ServiceClient.
+   * - If a `TokenCredential` was passed through the constructor, this `baseUri` defines the `getToken` scope to be `${options.baseUri}/.default`.
+   *
+   * If it is not specified:
+   * - All OperationSpecs must contain a baseUrl property.
+   * - If a `TokenCredential` was passed through the constructor, the `getToken` scope is set to be "https://management.azure.com/.default".
```
ramya-rao-a, line 170: Rewording to call out that baseUri set after the client is constructed does not get used by the adapter

```suggestion
   * The base URI against which requests will be made when using this ServiceClient instance.
   *
   * This can be set either by setting the `baseUri` in the `options` parameter to the ServiceClient constructor or directly after constructing the ServiceClient.
   * If set via the ServiceClient constructor when using the overload that takes the `TokenCredential`, this base URI sets the scope used to get the AAD token to `${baseUri}/.default` instead of the default "https://management.azure.com/.default"
   *
   * If it is not specified, all OperationSpecs must contain a baseUrl property.
```

ramya-rao-a:
That is a good catch regarding endpoint in the cognitive services packages @chradek 

> @chradek for the case you presented here: https://github.com/Azure/azure-sdk-for-js/blob/ce2c1f5310bf0e71dbf8bc2b29d5731a9dfd961f/sdk/cognitiveservices/cognitiveservices-entitysearch/src/entitySearchClientContext.ts#L40-L48

> I believe that scope is specific to the service version they use. For them, the default scope is likely sufficient, unless they’re working on external clouds.

@sadasant 

I am assuming that by "default scope" you mean `https://management.azure.com/.default`.
If so, then that will not work for non management plane packages that use `@azure/ms-rest-js`.
I had missed the fact that we have Track 1 data plane auto generated packages when deprioritizing the need for the adapter to respect the baseUri that has been set after the base class ServiceClient is constructed.

For example, I took a quick look at the newer Text Analytics package. Looks like the [baseUri is set to the endpoint](https://github.com/Azure/azure-sdk-for-js/blob/761a4e6692c3cd316a055c996188a3f68cb992ba/sdk/textanalytics/ai-text-analytics/src/generated/generatedClientContext.ts#L48-L50) and then in the core package the [`/.default` gets added to the baseUri to get the scopes](https://github.com/Azure/azure-sdk-for-js/blob/761a4e6692c3cd316a055c996188a3f68cb992ba/sdk/core/core-client/src/serviceClient.ts#L219-L228).

Therefore, we need to respect the protected property baseUri in the adapter that gets updated after ServiceClient is constructed even if a customer is using the default Azure cloud.

We have one of two options here
- Add a setter for the scope in the adapter so that we can set it to the right baseUri in each http request
- Keep changes in this PR, update the code generator to set the baseUri in the options before calling the base class constructor, regenerate all packages

ramya-rao-a:
@sadasant 

On second thoughts, none of the [older autogenerated cognitive services packages](https://github.com/Azure/azure-sdk-for-js/tree/ce2c1f5310bf0e71dbf8bc2b29d5731a9dfd961f/sdk/cognitiveservices) support AAD auth. So, one would not be trying to use the credentials from `@azure/identity` with any of them. We need to look for other non management plane packages that are auto generated to find a proper repro case :(

`@azure/batch` is a good candidate. I have logged https://github.com/Azure/azure-sdk-for-js/issues/16663 to update it to use credentials from `@azure/identity`. Once that is done, please work with @deyaaeldeen to test whether the changes done in this PR is enough to use the `@azure/batch` package in Azure public cloud

sadasant:
@ramya-rao-a sounds good! thank you!

ramya-rao-a:
On another set of second thoughts...

For other Azure clouds, the authority host needs to be set when creating the TokenCredential from `@azure/identity`. If the authority host was made a read only property on the credential, then can we not do something like the below in the adapter?

```js
switch (azureTokenCredential.authorityHost) {
      case "https://login.chinacloudapi.cn":
        this.scopes = "https://management.chinacloudapi.cn/.default"
        break;
      // and so on for other clouds
      default:
        break;
    }
```


This would break us away from relying on the baseUri which can be something totally different for non mgmt plane packages and not the right scope to use

ramya-rao-a:
@deyaaeldeen has been experimenting with the `@azure/batch` package and we concluded that we cannot reverse engineer the scope from the baseUri. Therefore, we are going to keep this `@azure/identity` support in Track 1 code gen only for the mgmt plane packages. This would mean that us hardcoding the mgmt related scope in the adapter in ms-rest-js is ok as long as we find a way to deal with different clouds. And so comes my suggestion in the previous comment on utilizing the authority host passed to the credential constructor to determine the right scope for mgmt plane packages

sadasant:
@ramya-rao-a Ok I’ll work on a PR for v1

sadasant:
@ramya-rao-a @deyaaeldeen on second thought

From the perspective of Identity, authorityHost does not fully overlap with scopes. Scopes are about what one wants to gain access to, and authorityHost is about who we want to authenticate with.

I think that’s part of the problem on ms-rest-js, baseUri is not a reliable scope for the same reason. It would be better if we could derive the scope during the authentication request, and not at the constructor.

ramya-rao-a:
Fair enough

In that case, lets guard the changes in this PR to only take affect if the baseUri matches the ARM scope
That way we atleast reduce the scope of this change to something we are sure should be updated.

We have logged https://github.com/Azure/autorest.typescript/issues/1153 to ensure that only the ARM packages get the TokenCredential support in the client constructor so that we don't have to worry what scope to use for data plane packages

sadasant:
I pushed something, but then I removed it from history.

I think there’s an issue with that proposal.

If I understand correctly, the proposed solution is to do the following:

- If I receive a `baseUri` that matches management, then we use a management scope that matches the Azure cloud that the `baseUri` has.
- However, the `baseUri` does not have a way to match any management request, since authentication endpoint !== scope. `baseUri` will for sure tell us what foreign cloud is being targeted, but it won’t say what resource clients want to get access to (which goes on scope).

For example: if `login.chinacloudapi.cn` is the `baseUri`, the scope can be `https://management.chinacloudapi.cn` or something else

sadasant:
I pushed a change that makes it so the baseUri is used as the scope only if it matches the management clouds. Please let me know if there’s anything else I can do!

lib/serviceClient.ts
```diff
@@ -185,9 +199,26 @@ export class ServiceClient {
...
+      const azureManagementClouds = [
+        "https://management.core.windows.net",
+        "https://management.core.chinacloudapi.cn",
+        "https://management.core.usgovcloudapi.net",
+        "https://management.core.cloudapi.de",
+      ];
+      if (
+        options?.baseUri &&
+        azureManagementClouds.find((cloud) => options!.baseUri!.indexOf(cloud) > -1)
+      ) {
```
ramya-rao-a, line 218: Can we simplify this?
```suggestion
      if (
        options?.baseUri &&
        azureManagementClouds.includes(options?.baseUri)
      ) {
```

Changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.6.0 - (2021-08-02)
```
ramya-rao-a, line 3:
```suggestion
## 2.6.0 - (2021-08-16)
```

lib/serviceClient.ts
sadasant: Includes breaks in the browser, I think https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

Well, InternetExplorer, so I guess it’s ok to use it

sadasant: I merged it, but keep in mind that includes will work only for exact matches.

ramya-rao-a: And that is good right? We want exact matches

sadasant: My brain is kinda fried right now. 🙇‍♂️ 🙇‍♂️ *slowly leaves*

Changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.6.0 - (2021-08-16)
+
+- Added a new property `baseUri` on the `ServiceClientOptions` that is then used to initialize the corresponding `baseUri` protected property on the `ServiceClient`.
+  - For `baseUri` that happen to be known Azure management clouds, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
```
ramya-rao-a, line 6:
```suggestion
  - For `baseUri` that happen to be known Azure resource management endpoints, this allows the instantiating of the `AzureIdentityCredentialAdapter` class with the right scope when a user constructs a `ServiceClient` with a `TokenCredential`. Resolves https://github.com/Azure/azure-sdk-for-js/issues/15945
```

lib/serviceClient.ts
```diff
@@ -185,9 +199,26 @@ export class ServiceClient {
...
 
+    if (options.baseUri) {
+      this.baseUri = options.baseUri;
+    }
+
     let serviceClientCredentials: ServiceClientCredentials | undefined;
     if (isTokenCredential(credentials)) {
-      serviceClientCredentials = new AzureIdentityCredentialAdapter(credentials);
+      let scope: string | undefined = undefined;
+      const azureManagementClouds = [
```
ramya-rao-a, line 209: Nit: This const can be defined outside of this class to avoid initialization every time a ServiceClient is created

ramya-rao-a (changes requested):

lib/serviceClient.ts
ramya-rao-a: As far as we know, the baseuri indeed needs to match the resource manager endpoints we have listed here. If there are cases where this is not true, I would much rather hear that from a customer than make assumptions ourselves

ramya-rao-a:
Another nit: Is the term `resource management endpoint` preferred over `resource manager endpoints`?
https://github.com/Azure/ms-rest-azure-env/blob/master/lib/azureEnvironment.ts#L271 uses the latter

And official docs refer to the latter as well. See https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/overview

lib/serviceClient.ts
sadasant: Addressed here: https://github.com/Azure/ms-rest-js/pull/457/commits/7224651161fbf7928a08f861a864d5dd34064ac1

lib/serviceClient.ts
sadasant: Addressed here: https://github.com/Azure/ms-rest-js/pull/457/commits/7224651161fbf7928a08f861a864d5dd34064ac1

sadasant:
@ramya-rao-a I think I’ve addressed all the feedback through this commit: https://github.com/Azure/ms-rest-js/pull/457/commits/7224651161fbf7928a08f861a864d5dd34064ac1

lib/serviceClient.ts
```diff
@@ -143,6 +146,16 @@ export interface ServiceClientOptions {
...
    */
   agentSettings?: AgentSettings;
+  /**
+   * If specified, and if the value matches a known resource manager endpoint:
+   * - This `baseUri` becomes the base URI that requests will be made against for this ServiceClient.
+   * - If a `TokenCredential` was passed through the constructor, this `baseUri` defines the `getToken` scope to be `${options.baseUri}/.default`.
+   *
+   * If it is not specified, or if it doesn't match a known resource manager endpoint:
+   * - All OperationSpecs must contain a baseUrl property.
+   * - If a `TokenCredential` was passed through the constructor, the `getToken` scope is set to be "https://management.azure.com/.default".
```
ramya-rao-a, line 156:
```suggestion
   * If specified:
   * - This `baseUri` becomes the base URI that requests will be made against for this ServiceClient.
   * - If the `baseUri` matches a known resource manager endpoint and if a `TokenCredential` was passed through the constructor, this `baseUri` defines the `getToken` scope to be `${options.baseUri}/.default`. Otherwise, the scope would default to "https://management.azure.com/.default".
   *
   * If it is not specified:
   * - All OperationSpecs must contain a baseUrl property.
   * - If a `TokenCredential` was passed through the constructor, the `getToken` scope is set to be "https://management.azure.com/.default".
```

Changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.6.0 - (2021-08-16)
```
ramya-rao-a, line 3:
```suggestion
## 2.6.0 - (2021-08-18)
```

ramya-rao-a (changes requested):

ramya-rao-a (approved):

PR #460, opened 2022-01-24
Fix build pipelines

jeremymeng:
Update vmImage to use Ubuntu 18.04 as 16.04 has been removed from devops.


jeremymeng:
Check_Everything failed because of version check. We don't need a version bump for this change.

xirzec (approved):

PR #464, opened 2022-02-14
Add stub interfaces

jeremymeng:
so we can stop depending on @types/node


jeremymeng:
By adding stubs with a combination of d.ts files and interfaces.ts. I no longer see "node" reference in our generated d.ts files, nor our api.md review file.

tsconfig.es.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "module": "es6",
-    "outDir": "es"
+    "outDir": "es",
+    "types": ["express"]
```
jeremymeng, line 6: I tried this but it doesn't seem to stop "node" from being visible in compilation.

